PR #6, opened 2023-10-20
Fix CI

WashingtonKK:


.github/workflows/main.yaml
```diff
@@ -29,8 +29,7 @@ jobs:
 
       - name: Build
         run: |
-          go build -mod=vendor -ldflags="-linkmode=external -extldflags=-static -s -w" cmd/agent/main.go
-          go build -mod=vendor -ldflags "-s -w" cmd/cli/main.go
+          go build -v ./...
```
drasko, line 32: Is there a particular reason why you are changing linker flags?

drasko (changes requested):

.github/workflows/main.yaml
WashingtonKK: No, I replicated what was done on prism.

drasko (changes requested):

.github/workflows/main.yaml
```diff
@@ -29,7 +29,15 @@ jobs:
 
       - name: Build
         run: |
-          go build -v ./...
+          make agent
```
drasko, line 32: There is no particular reason to make service by service in CI - please use just `make` to make all services at once.

drasko (changes requested):

.github/workflows/main.yaml
WashingtonKK: It is fine now

drasko (approved):
Approving, but we should consider to add `test` target to `Makefile` as well. That would be the most elegant.

PR #19, opened 2023-11-14
NOISSUE - Add CoCos-AI Manager API specification

SammyOina:
Add openapi spec


api/manager.yml
```diff
@@ -0,0 +1,145 @@
...
+          example: 7d6f4dc4f7f0c1fa3dc24eddfb18bb5073ff4f62
+        description:
+          type: string
+          description: Service description.
+          example: <service_name> service
+        build_time:
+          type: string
+          description: Service build time.
+          example: 1970-01-01_00:00:00
+    ComputationReq:
```
rodneyosodo, line 71: This is declared but not used

rodneyosodo (changes requested):

api/manager.yml
rodneyosodo: But it is currently not used https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it

SammyOina: it is referred to in the spec, the bytes in the request are from the /run request arise from this

rodneyosodo: From the documentation, there is nothing linking the two. Something like 
```yaml
description: byte array of computation request generated from ...
```

SammyOina: see l130

drasko (approved):
LGTM

PR #23, opened 2023-11-24
NOISSUE - Add linters

SammyOina:
Add linters
- switchTrue
- importShadow
-  httpNoBody 
- paramTypeCombine 
- emptyStringTest  
- builtinShadow  
- exposedSyncMutex  
- importas

### DIFF
--- a/internal/libvirt/libvirt.go
+++ b/internal/libvirt/libvirt.go
@@ -10,12 +10,12 @@ import (
 	"regexp"
 	"strings"
 
-	"github.com/digitalocean/go-libvirt"
+	golibvirt "github.com/digitalocean/go-libvirt"
 )
 
 var re = regexp.MustCompile(`'([^']*)'`)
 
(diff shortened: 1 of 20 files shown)


internal/libvirt/libvirt.go
```diff
@@ -10,12 +10,12 @@ import (
 	"regexp"
 	"strings"
 
-	"github.com/digitalocean/go-libvirt"
+	golibvirt "github.com/digitalocean/go-libvirt"
```
rodneyosodo, line 13: Why import aliasing for this? If it is necessary 
```suggestion
	libvirt "github.com/digitalocean/go-libvirt"
```

rodneyosodo (changes requested):

internal/libvirt/libvirt.go
SammyOina: I use golibvirt to avoid an import shadow

rodneyosodo (approved):
LGTM

drasko (approved):
LGTM

PR #32, opened 2024-01-02
NOISSUE - Optimized Docker build, split into stages

SammyOina:
Refactored the Dockerfile to use multi-stage builds, improving the image size and separation of concerns. Removed `qemu-system-x86_64` from the initial build stage and included it in a new, minimal final image based on `alpine:latest`.


rodneyosodo (approved):
LGTM

dborovcanin (approved):

PR #40, opened 2024-01-08
Use systemd for HAL boot

danko-miladinovic:
The `cocos_config` file has been changed to use `systemd` for HAL boot. `Agent` configuration has also been modified in order for it to run on startup.


SammyOina (approved):

hal/linux/configs/cocos_defconfig
```diff
@@ -4,7 +4,9 @@ BR2_x86_64=y
 # System
 BR2_TARGET_GENERIC_HOSTNAME="cocos"
 BR2_TARGET_GENERIC_ISSUE="Welcome to Cocos"
-BR2_SYSTEM_DHCP="eth0"
+BR2_SYSTEM_DHCP="enp0s2"
```
drasko, line 7: Remove dynamic address setup, define static IP addr config, call iface `eth0`

drasko (changes requested):

hal/linux/configs/cocos_defconfig
drasko: Make it `10.0.0.1`

init/systemd/00-internal0.link
```diff
@@ -0,0 +1,6 @@
+[Match]
+OriginalName=*
+
+[Link]
+Name=internal0
```
drasko, line 5: What is this `internal0`, and why a name like this was selected? Is there some standard naming for systemd?

hal/linux/package/agent/agent.mk
```diff
@@ -12,7 +12,14 @@ define AGENT_BUILD_CMDS
...
+	$(INSTALL) -D -m 0755 $(@D)/build/cocos-agent $(TARGET_DIR)/bin
+	mkdir -p $(TARGET_DIR)/var/log/cocos
 endef
 
-$(eval $(golang-package))
\ No newline at end of file
+define AGENT_INSTALL_INIT_SYSTEMD
+	$(INSTALL) -D -m 0644 $(@D)/init/systemd/cocos-agent.service $(TARGET_DIR)/usr/lib/systemd/system/cocos-agent.service
+	$(INSTALL) -D -m 0644 $(@D)/init/systemd/eth0.network $(TARGET_DIR)/etc/systemd/network
+	$(INSTALL) -D -m 0644 $(@D)/init/systemd/00-internal0.link $(TARGET_DIR)/etc/systemd/network
```
drasko, line 22: I am not sure that you need to prefix this with `00-`. How is `systemd` establishing the order of service / init script deployment? I think you need to use `After=` or similar directives: https://serverfault.com/questions/482730/systemd-dependencies-and-boot-order, https://www.digitalocean.com/community/tutorials/understanding-systemd-units-and-unit-files

drasko (changes requested):

hal/linux/package/agent/agent.mk
```diff
@@ -12,7 +12,14 @@ define AGENT_BUILD_CMDS
...
+	$(INSTALL) -D -m 0755 $(@D)/build/cocos-agent $(TARGET_DIR)/bin
+	mkdir -p $(TARGET_DIR)/var/log/cocos
 endef
 
-$(eval $(golang-package))
\ No newline at end of file
+define AGENT_INSTALL_INIT_SYSTEMD
+	$(INSTALL) -D -m 0644 $(@D)/init/systemd/cocos-agent.service $(TARGET_DIR)/usr/lib/systemd/system/cocos-agent.service
+	$(INSTALL) -D -m 0644 $(@D)/init/systemd/eth0.network $(TARGET_DIR)/etc/systemd/network
+	$(INSTALL) -D -m 0644 $(@D)/init/systemd/00-network.link $(TARGET_DIR)/etc/systemd/network
```
danko-miladinovic, line 22: I have changed the name of the .link file.

There are a couple of ways for a network interface to get its name:

1. By kernel - this can be unpredictable if there are multiple network interfaces because the kernel does device probing in random order and thus can give different names to network interfaces during reboots. For example, the first ethernet device detected gets the name `eth0`, the second `eth1`, and so on. We have only one interface, so it will be named `eth0`.
2. By `systemd` - the `systemd` gives predictable names and provides them based on their location in the system. 
3. By user - using ip command or some other way.

The name of the file starts with numeric numbers in order for it to be processed by the system before the default network configuration file. From the documentation:
```The link file contains a [Match] section, which determines if a given link file may be applied to a given device, as well as a [Link] section specifying how the device should be configured. The first (in lexical order) of the link files that matches a given device is applied. Note that a default file 99-default.link is shipped by the system. Any user-supplied .link should hence have a lexically earlier name to be considered at all.```

File: init/systemd/00-network.link [hunk 5]
```diff
@@ -0,0 +1,6 @@
+[Match]
+OriginalName=*
+
+[Link]
+NamePolicy=kernel
```
danko-miladinovic, line 5: The name policy here will keep the kernel name because there is only one network ethernet interface, so it is consistent.

drasko: What does this `NamePolicy=kernel` actually means?

drasko: @danko-miladinovic would a super-simple file like this work:
```bash
[Link]
Name=eth0
```

Do we need part
```bash
[Match]
OriginalName=*
```

drasko: Kernel can give other names - like enp0s3` for example.

I think it would be safer to use:
```bash
[Match]
OriginalName=*

[Link]
Name=eth0
```

but why are we doing all of this? We should fetch this info (which iface to use) from Qemu command line, right?

drasko (changes requested):

PR #43, opened 2024-01-09
COCOS-25 - Remove networking between manager and agent

### LINKED ISSUE #25: Remove networking between Manager and Agent
There is no need for Manager to communicate with the Agent directly. Pass Computation Manifest in some other way, then do not enable local network communication. Manager needs only to monitor SVM on the OS level, not to talk to the internal SVM apps. Agent needs to communicate directly with the Computations service by sending notifications (as a client).

SammyOina:
Resolves #25 
To be merged after #40  and #42


drasko (approved):
LGTM

PR #54, opened 2024-01-15
Use DHCP for VM networking

danko-miladinovic:
DHCP has been added for VM networking. Key changes:

- File `eth0.network` has been deleted. This file was used to set a static ip address. The 00-network.link file has not been deleted in order for the ethernet network interface name to stay `eth0`.
- Script `agent_start_script.sh` has been added in order to fetch the ip address of the VM and to call `dhclient eth0` (for the same reason).
- Files `cocos_defconfig` and `agent.mk` have been adjusted for this whole process to work.


init/systemd/agent_start_script.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+dhclient eth0
+AGENT_GRPC_HOST=$(ip -4 addr show eth0 | grep inet | awk '{print $2}' | cut -d/ -f1)
```
drasko, line 4: You kept name eth0?

drasko (changes requested):

init/systemd/agent_start_script.sh
danko-miladinovic: I have changed the name so it uses systemd. It is now `enp0s2`.

init/systemd/agent_start_script.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+ip link set dev enp0s2 up
+dhclient enp0s2
+AGENT_GRPC_HOST=$(ip -4 addr show enp0s2 | grep inet | awk '{print $2}' | cut -d/ -f1)
```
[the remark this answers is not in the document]
danko-miladinovic: Done

init/systemd/agent_start_script.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+ETH_IFACE=enp0s2
```
drasko, line 3: Put a comment here please why this name, and explain that it should match Qemu params

drasko (changes requested):
LGTM

init/systemd/agent_start_script.sh
danko-miladinovic: I have added an explanation for the `ETH_IFACE` name.

drasko (changes requested):
LGTM

PR #57, opened 2024-01-16
COCOS-49 - Pass agent configuration and computation via vsock

### LINKED ISSUE #49: Add agent configuration on vsock
https://github.com/ultravioletrs/cocos/blob/200f20767319138dc65e49a7a13387a73084ddf1/init/systemd/cocos-agent.service#L9C37-L11

Agent needs notifications URL in the environment. This will need to be defined when building the image (HAL).

SammyOina:
Resolves #49 
Requires #61 
- Added ability to read agent configurations directly from virtual socket connections.
- Streamlined computation configuration by removing outdated structures and adopting a new data model.
- Updated main agent application to utilize new configuration reading method.
- Adjusted manager service logic to align with the new computation and agent configuration model.


cmd/agent/main.go
```diff
@@ -111,30 +97,29 @@ func newService(ctx context.Context, logger mglog.Logger, eventSvc events.Servic
 	return svc
 }
 
-// extractComputationValue to extract computation value from the command line.
-func extractComputationValue() (agent.Computation, error) {
-	cmdLineBytes, err := os.ReadFile("/proc/cmdline")
+func readConfig() (agent.Computation, error) {
+	l, err := vsock.Listen(manager.ManagerPort, nil)
```
drasko, line 101: Why `l`? This is confusing. Why not `vs` if this is vsocket?

drasko (changes requested):

cmd/agent/main.go
SammyOina: `l` for listener

drasko: Yep, all clear - thanks.

drasko (approved):
LGTM

PR #63, opened 2024-01-18
COCOS-56 - Pass agent computation events to manager

SammyOina:
Resolves #56 

To be merged after #57

PR #64, opened 2024-01-18
COCOS-55 - Pass agent logs to manager via vsock

### LINKED ISSUE #55: Pass agent logs via vsock to manager

SammyOina:
Enhance agent logging via vsock connection

Redirected agent logging to use a vsock connection instead of standard output, improving the process isolation and enabling centralized log management. The change involved dialing to the specified vsock log port and initializing the logger with the vsock connection rather than stdout.

Additionally, the manager service now maintains a map of agent vsock cids to computation IDs, providing better tracking of computation resources. A routine to retrieve logs from agents was also initiated during the service setup to facilitate log collection.

Consequential to these changes is the removal of a redundant os package import in the agent's main.go, further cleaning up the dependencies.

Update to latest magistrala version

Resolves #55 
To be merged after #63


manager/agentEvents/agentEvents.go
drasko, line 1: Do not use camelCase for directory and package names - it is not Go idiomatic. Consider simplifying the name - for example just `agent` or `events` might work, when within the `manager` dir. Maybe even `agentlog` or eve `agentcomm` if this is a bidirectional communication (if the same vsock and functions in this package will be used for sending params and other messages from Manager to Agent). Otherwise, you can rename directory to `agentevents`.

drasko (changes requested):

drasko (approved):
LGTM

PR #66, opened 2024-01-24
NOISSUE - Fetch attestation report

SammyOina:
Enhanced the attestation process to require additional report data and a variable privacy level, catering to varying security requirements. The `Run` method's context parameter was removed to simplify invocation following API alignment. Dependencies got updated, notably adding `github.com/google/go-sev-guest` for attestation, and some were indirectly bumped. These updates ensure the agent service leverages the latest available features and improvements for security and stability.


agent/agent.proto
```diff
@@ -36,6 +36,9 @@ message ResultRequest {
 
 message ResultResponse { bytes file = 1; }
 
-message AttestationRequest { }
+message AttestationRequest {
+  bytes report_data = 1;
+  uint32 vpml = 2;
```
danko-miladinovic, line 41: We should remove `vmpl` as a parameter. VMPL stands for virtual machine privilege level and is a mechanism to further increase security in the VM by dividing its address space into four levels. Leve 0, VMPL0, is the most privileged one, and VMPL3 is the least. We do not support VMPL, and in our case, everything is running inside level 0. 

This mechanism can be used, and is used, in the support of virtual TPMs (vTPM). vTPM is used for security purposes and is run in VMPL0, and the guest OS is run in VMPL3 (or any other level).

agent/service.go
```diff
@@ -195,14 +196,23 @@ func (as *agentService) Result(ctx context.Context, consumer string) ([]byte, er
 	return as.result, as.runError
 }
 
-func (as *agentService) Attestation(ctx context.Context) ([]byte, error) {
-	// Implement the logic for the Attestation method here
-	// Use the provided ctx parameter as needed
-	var attestation []byte
+func (as *agentService) Attestation(ctx context.Context, reportData []byte, vpml uint) ([]byte, error) {
+	device, err := client.OpenDevice()
```
danko-miladinovic, line 200: `client.OpenDevice()` is not needed. The method `GetRawQuoteAtLevel` will open the device, fetch the attestation report and close it.

danko-miladinovic:
LGTM. The attestation report can be fetched from the agent.

WashingtonKK (approved):
LGTM

drasko (approved):
LGTM

PR #71, opened 2024-01-29
NOISSUE - mTLS support across services

SammyOina:
Extended gRPC configuration to support mutual TLS (mTLS) in agent and manager components for enhanced security. This includes the loading of Certificate Authority (CA) certificates, server, and client certificates, and keys. Updated README documentation to reflect the new environment variables required for mTLS configuration. Additionally, streamlined secure gRPC client connection setup and logging messages to indicate whether a service is running with TLS, mTLS, or without TLS.

The change ensures secure communication between services by verifying both client and server identities, thus addressing potential security concerns in network-level interactions.


internal/server/grpc/grpc.go
```diff
@@ -102,3 +146,36 @@ func (s *Server) Stop() error {
 
 	return nil
 }
+
+func loadCertFile(certFile string) ([]byte, error) {
+	if certFile != "" {
+		return os.ReadFile(certFile)
+	}
+	return []byte{}, nil
+}
```
WashingtonKK, line 155: Why not just call os.ReadFile instead of having this function?
os.ReadFile will return an error when the cert is missing, which you can check directly.

SammyOina: this function is used multiple times and has more than one check, using this function saves on locs

arvindh123 (commented):
LGTM

arvindh123 (approved):

drasko (approved):
LGTM

PR #85, opened 2024-02-19
NOISSUE - Simplify agent config

SammyOina:
Centralized agent's GRPC port definition by removing the host and port fields from multiple configurations and using a single source of truth from the QEMU manager. This change simplifies the networking setup and avoids redundancy by ensuring the agent's GRPC service port is defined consistently across the application. Adjustments were made across configuration structs, protobuf definitions, and service instantiation to reflect this new approach.


drasko:
Let's leave this flexibility of host/port config for Cocos users

PR #86, opened 2024-02-22
NOISSUE - Enforce 32-byte length check for hashes

SammyOina:
Introduced a validation step in the manager service to ensure both algorithm and dataset hashes are of byte length 32 before provisioning VMs, preventing runtime errors due to invalid hash lengths. The test manager-server now generates a valid 32-byte hash using SHA256 for mock data to align with the new validation requirement.

### DIFF
--- a/manager/service.go
+++ b/manager/service.go
@@ -18,6 +18,8 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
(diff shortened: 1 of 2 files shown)


drasko (changes requested):

drasko (approved):
LGTM

PR #87, opened 2024-02-26
NOISSUE - Add attesatation validation/verification options

SammyOina:
- **New Features**
	- Added functionality to retrieve and save attestation information with configurable validation parameters.
	- Enhanced validation process with new options like checking CRL, disallowing network downloads, setting timeouts, and more.
- **Refactor**
	- Updated attestation data handling to use a fixed-length byte array for consistency and validation.
	- Modified method signatures in logging and metrics middleware for improved data handling.
- **Bug Fixes**
	- Improved attestation request and validation logic for accurate processing and verification.
- **Documentation**
	- Updated internal documentation to reflect changes in attestation data handling and configuration options.


cli/README.md
```diff
@@ -21,16 +21,49 @@ To run a computation, use the following command:
...
 ```
 
 #### Validate attestation
+Validates the retrieved attestation information against a specified policy and checks its authenticity.
 To validate and verify attestation from agent, use the following command:
 ```bash
-./build/cocos-cli agent attestation validate '<attestation>' '<report_data>'
+./build/cocos-cli agent attestation validate '<attestation>' --report_data '<report_data>'
 ```
+##### Flags
```
danko-miladinovic, line 36: Can we add an option for `guest_policy`? Like here:

https://github.com/google/go-sev-guest/blob/main/tools/check/check.go

SammyOina: we do have it as `guest_svn` https://github.com/ultravioletrs/cocos/blob/b0a81d400ca9a235fdf5f837a61df7af10c3a4c0/cli/attestation.go#L174

cli/attestation.go
```diff
@@ -70,32 +120,261 @@ func (cli *CLI) NewGetAttestationCmd() *cobra.Command {
...
+	}
 	attestationPB, err := abi.ReportCertsToProto(attestation)
 	if err != nil {
 		return err
 	}
-	if err = verify.SnpAttestation(attestationPB, verify.DefaultOptions()); err != nil {
+	if err = verify.SnpAttestation(attestationPB, sopts); err != nil {
+		return err
+	}
+	opts, err := validate.PolicyToOptions(cfg.Policy)
```
danko-miladinovic, line 218: This check returns an error. Field `cfg.Policy.Policy` should be initialized or set to default value. The default value is: `0x0000000000030000`. 

The value for `cfg.Policy.Policy` should be initialized by the option `guest_policy` or with the above mentioned default value.

cli/attestation.go
```diff
@@ -70,32 +120,261 @@ func (cli *CLI) NewGetAttestationCmd() *cobra.Command {
 }
 
 func (cli *CLI) NewValidateAttestationValidationCmd() *cobra.Command {
```
danko-miladinovic, line 122: The following fields should have default values:
- `host_data` - default value should be `0` (the field should be filled with zeros).
- `family_id` - default value should be `0` (the field should be filled with zeros).
- `image_id` - default value should be `0` (the field should be filled with zeros).
- `report_id` - default value should be `0` (the field should be filled with zeros).
- `report_id_ma` - default value should be `0` (the field should be filled with zeros).

cli/attestation.go
```diff
@@ -70,32 +120,261 @@ func (cli *CLI) NewGetAttestationCmd() *cobra.Command {
...
+		return err
+	}
+	if err := validateFieldLength("report_id_ma", cfg.Policy.ReportIdMa, size32); err != nil {
 		return err
 	}
-	if err = validate.SnpAttestation(attestationPB, &validate.Options{ReportData: reportData}); err != nil {
+	if err := validateFieldLength("measurement", cfg.Policy.Measurement, size48); err != nil {
 		return err
 	}
+	if err := validateFieldLength("chip_id", cfg.Policy.ChipId, size48); err != nil {
```
danko-miladinovic, line 359: The length of the filed `chip_id` should be 64 bytes. As per the documentatio:
https://www.amd.com/content/dam/amd/en/documents/epyc-technical-docs/specifications/56860.pdf

**`Page 46`**

cli/README.md
danko-miladinovic: Ok, I got confused for a moment because there is a  field in the attestation report called Guest SVN, which is not guest policy and the value for guest policy should initialize `cfg.Policy.Policy`.

agent/api/logging.go
```diff
@@ -66,7 +66,7 @@ func (lm *loggingMiddleware) Result(ctx context.Context, consumer string) (respo
 	return lm.svc.Result(ctx, consumer)
 }
 
-func (lm *loggingMiddleware) Attestation(ctx context.Context, reportData []byte) (response []byte, err error) {
+func (lm *loggingMiddleware) Attestation(ctx context.Context, reportData [64]byte) (response []byte, err error) {
```
drasko, line 69: Magic numbers?

drasko (changes requested):

cli/README.md
```diff
@@ -21,16 +21,49 @@ To run a computation, use the following command:
...
+- --platform_info: Maximum acceptable PLATFORM_INFO field bit-wise (optional).
+- --minimum_version: Minimum AMD-SP firmware API version (major.minor) (optional).
+- --trusted_author_keys: Paths to x.509 certificates of trusted author keys (optional).
+- --trusted_author_key_hashes: Hex-encoded SHA-384 hashes of trusted author keys (optional).
+- --trusted_id_keys: Paths to x.509 certificates of trusted identity keys (optional).
+- --trusted_id_key_hashes: Hex-encoded SHA-384 hashes of trusted identity keys (optional).
+- --product: AMD product name for the chip that generated the attestation report (optional).
+- --stepping: Machine stepping for the chip that generated the attestation report (optional).
+- --CA_bundles_paths: Paths to CA bundles for the AMD product (optional).
+- --CA_bundles: PEM format CA bundles for the AMD product (optional).
```
danko-miladinovic, line 66: Please add that the user must specify the `product` option when using the `CA_bundles_paths` and /or `CA_bundles` option.

agent/api/grpc/client.go
```diff
@@ -208,7 +208,7 @@ func (c grpcClient) Attestation(ctx context.Context, request *agent.AttestationR
 	ctx, cancel := context.WithTimeout(ctx, c.timeout)
 	defer cancel()
 
-	res, err := c.attestation(ctx, &attestationReq{ReportData: request.ReportData})
+	res, err := c.attestation(ctx, &attestationReq{ReportData: [64]byte(request.ReportData)})
```
drasko, line 211: Define this `64` somewhere

agent/api/logging.go
```diff
@@ -66,7 +66,7 @@ func (lm *loggingMiddleware) Result(ctx context.Context, consumer string) (respo
 	return lm.svc.Result(ctx, consumer)
 }
 
-func (lm *loggingMiddleware) Attestation(ctx context.Context, reportData []byte) (response []byte, err error) {
+func (lm *loggingMiddleware) Attestation(ctx context.Context, reportData [size64]byte) (response []byte, err error) {
```
drasko, line 69: what is this constant `size64`? Where does it come from?

drasko (changes requested):

agent/api/logging.go
SammyOina: report data is 64 bytes long, https://github.com/google/go-sev-guest/blob/08d92a6f9da9dcab1ca702485aff381b5d3efdf9/proto/sevsnp.proto#L38

drasko: This is OK, but is this some constant, and can we please declare it then at the beginning of the file, like `att_report_size` or something like this

.github/workflows/checkproto.yaml
```diff
@@ -55,7 +55,7 @@ jobs:
       - name: Set up Cocos-AI
         run: |
           # Rename .pb.go files to .pb.go.tmp to prevent conflicts
-          for p in $(ls manager/*.pb.go); do
+          for p in $(ls pkg/manager/*.pb.go); do
```
drasko, line 58: Why is this package called exactly as the `manager` package at the top level? This does not correspond to how we do things in Magistrala, and adds confusion.

@dborovcanin I think this name in `pkg` should be changed to something else, right? Maybe `managersdk` or something else.

drasko (changes requested):

.github/workflows/checkproto.yaml
SammyOina: the files in pkg/manager are the generated proto files. I moved them there since they are imported in prism/computations, manager/api/grpc and manager/manager. manager/api/grpc and manager/manager is also imported in prism/computations. I moved the generated files to /pkg/manager from manager/manager to avoid circular import

agent/api/logging.go
SammyOina: we use it a number of times so I'll set an exported constant

.github/workflows/checkproto.yaml
drasko: It does not matter how files are generated, but how the directory is called. Can we please change the naming, because it is confusing?

drasko: @dborovcanin please help here urgently - decide naming of the directory and help with the code structure for imports

cli/README.md
dborovcanin: Is this resolved (`guest_policy` field is not added as far as I can see).

cli/attestation.go
```diff
@@ -70,32 +123,265 @@ func (cli *CLI) NewGetAttestationCmd() *cobra.Command {
...
+				log.Fatalf("attestation validation and verification failed with error: %s", err)
+			}
+
+			if err := verifyAndValidateAttestation(attestation); err != nil {
+				log.Fatalf("attestation validation and verification failed with error: %s", err)
 			}
 			log.Println("Attestation validation and verification is successful!")
 		},
 	}
+	cmd.Flags().StringVar(&cfgString, "config", "", "Serialized json check.Config protobuf. This will overwrite individual flags. Unmarshalled as json. Example: "+`
```
dborovcanin, line 165: Move the example to the separate constant and use pretty JSON formatting.

cli/attestation.go
```diff
@@ -70,32 +123,265 @@ func (cli *CLI) NewGetAttestationCmd() *cobra.Command {
...
+	cmd.Flags().BytesHexVar(&cfg.Policy.HostData, "host_data", empty32[:], "The expected HOST_DATA field as a hex string. Must encode 32 bytes. Unchecked if unset.")
```
dborovcanin, line 167: Use multiline, it helps readability:
```go
cmd.Flags().BytesHexVar(
	&cfg.Policy.HostData,
	"host_data",
	empty32[:],
	"The expected HOST_DATA field as a hex string. Must encode 32 bytes. Unchecked if unset.",
)
```

dborovcanin (changes requested):

.github/workflows/checkproto.yaml
SammyOina: @dborovcanin this is the circular import I was referring to https://github.com/ultravioletrs/cocos/pull/87#discussion_r1528391334

dborovcanin (approved):

PR #94, opened 2024-03-12
NOISSUE - Implement manager client heartbeat mechanism 

SammyOina:
Introduced a heartbeat system in the manager service to maintain an active link with client agents. The updates involve sending periodic heartbeat signals controlled via the new environment variable `MANAGER_HEARTBEAT_INTERVAL`, defaulting to 1 second. The protobuf definitions, service interfaces, and server logic have been updated accordingly to handle the incoming signals. This change allows better tracking of active client connections and could be used for features like auto-reconnect or resource cleanup for lost connections in the future.


cmd/manager/main.go
```diff
@@ -107,6 +110,16 @@ func main() {
 		logger.Error(err.Error())
 		return
 	}
+	hc, err := managerClient.Heartbeat(ctx)
+	if err != nil {
+		logger.Error(err.Error())
+		return
+	}
+	hbs := heartbeat.New(hc, cfg.HeartbeatInterval)
```
drasko, line 118: Why `hbs` and not `hb`?

manager/api/grpc/server.go
```diff
@@ -60,3 +64,16 @@ func (s *grpcServer) Process(stream manager.ManagerService_ProcessServer) error
...
 	return eg.Wait()
 }
+
+func (s *grpcServer) Heartbeat(stream manager.ManagerService_HeartbeatServer) error {
+	p, ok := peer.FromContext(stream.Context())
+	if !ok {
+		return errors.New("failed to get peer from context")
+	}
+	for {
+		if _, err := stream.Recv(); err != nil {
```
drasko, line 74: What is this that we are receiving? I thought that in the `Heartbeat ()` fnc we have just to send a heartbeat message.

drasko (changes requested):

manager/api/grpc/server.go
SammyOina: this on the server side where heartbeat requests are received

drasko:
Closed in the favor of https://github.com/ultravioletrs/cocos/issues/104

PR #95, opened 2024-03-14
NOISSUE - Allow termination of manager

SammyOina:
# What type of PR is this?

This is a feature because it adds manager client termination.

## What does this do?

Allows server to send termination requests to clients
Provides server with authentication information used by server

- None

No, I have tested manually

No, I have not updated the documentation because no change in outward functionality.

### Notes
to be merged after #94


drasko:
I would need @dborovcanin and @arvindh123 to review his one

.github/workflows/checkproto.yaml
```diff
@@ -55,7 +55,7 @@ jobs:
       - name: Set up Cocos-AI
         run: |
           # Rename .pb.go files to .pb.go.tmp to prevent conflicts
-          for p in $(ls manager/*.pb.go); do
+          for p in $(ls pkg/manager/*.pb.go); do
```
drasko, line 58: I do not like that this in `pkg` is called `manager`, like other `manager` in the top level. It is confusing.

drasko (changes requested):

test/computations/main.go
```diff
@@ -53,17 +54,21 @@ func (s *svc) Run(ipAdress string, reqChan chan *manager.ComputationRunReq) {
...
-		ResultConsumers: []string{"consumer1"},
-		AgentConfig: &manager.AgentConfig{
-			Port:        "7002",
-			LogLevel:    "debug",
-			AttestedTls: attestedTLS,
+	reqChan <- &manager.ServerStreamMessage{
+		Message: &manager.ServerStreamMessage_RunReq{
+			RunReq: &manager.ComputationRunReq{
+				Id:              "1",
+				Name:            "sample computation",
```
drasko, line 61: What is this naming (`sample ...`)? Looks like some test, but it is actually in `main()`

drasko (changes requested):

test/computations/main.go
SammyOina: this is the test server which we use as an example for users as a standin for backends service

drasko (approved):
LGTM

PR #96, opened 2024-03-15
NOISSUE - Fix network interface problem for agent

danko-miladinovic:
# What type of PR is this?
This is a fix for agent networking interface problem and removes SCSI parameter from the QEMU command (or SCSI interface from the VM).

## What does this do?
The problem was that the agent could not connect to the network, because the network interface did not have the default (expected) name. The name is given to the network interface by `systemd` based on the network type (ethernet, WLAN etc.), slot number etc. 

Also, the `virtio-scsi-pci` parameter is removed from the QEMU command, because we no longer have a virtual disk for the VM.

No issue

The Manager and the Manager test server should be up and running (look at the manager README file). When the VM is started, the `cli` tool should be used.

Documentation will be updated when the whole flow is tested.

### DIFF
--- a/manager/qemu/config.go
+++ b/manager/qemu/config.go
@@ -37,6 +37,8 @@ type NetDevConfig struct {
 type VirtioNetPciConfig struct {
 	DisableLegacy string `env:"VIRTIO_NET_PCI_DISABLE_LEGACY" envDefault:"on"`
 	IOMMUPlatform bool   `env:"VIRTIO_NET_PCI_IOMMU_PLATFORM" envDefault:"true"`
+	Bus           string `env:"VIRTIO_NET_PCI_BUS" envDefault:"pcie.0"`
+	Addr          string `env:"VIRTIO_NET_PCI_ADDR" envDefault:"0x2"`
 	ROMFile       string `env:"VIRTIO_NET_PCI_ROMFILE"`
 }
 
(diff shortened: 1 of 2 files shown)


SammyOina (approved):

SammyOina (approved):

SammyOina (approved):

manager/qemu/config.go
```diff
@@ -140,36 +134,30 @@ func constructQemuArgs(config Config) []string {
...
-		fmt.Sprintf("virtio-net-pci,disable-legacy=%s,iommu_platform=%v,netdev=%s,romfile=%s",
+		fmt.Sprintf("virtio-net-pci,disable-legacy=%s,iommu_platform=%v,netdev=%s,addr=%s,romfile=%s",
 			config.VirtioNetPciConfig.DisableLegacy,
 			config.VirtioNetPciConfig.IOMMUPlatform,
 			config.NetDevConfig.ID,
+			config.VirtioNetPciConfig.Addr,
 			config.VirtioNetPciConfig.ROMFile))
 
+	args = append(args, "-device", fmt.Sprintf("vhost-vsock-pci,id=%s,guest-cid=%d", config.VSockConfig.ID, config.VSockConfig.GuestCID))
+
```
drasko, line 152: Remove these blank lines

drasko (changes requested):

drasko (approved):
LGTM

PR #99, opened 2024-03-31
NOISSUE  - Attested TLS

danko-miladinovic:
# What type of PR is this?
This is a feature because it adds the functionality of Attested TLS.

## What does this do?
This PR adds the attestation report to a self-signed x.509 certificate. The CLI application then verifies this custom extension.

No issue.

The instructions for manual testing have been updated.

The documentation will be added in separate pull request.

### Notes
The file `computation.json` should be used during verification. It contains the necessary information for attestation verification.


cmd/agent/main.go
```diff
@@ -66,13 +66,14 @@ func main() {
 		KeyFile:      cfg.AgentConfig.KeyFile,
 		ServerCAFile: cfg.AgentConfig.ServerCAFile,
 		ClientCAFile: cfg.AgentConfig.ClientCAFile,
+		AttestedTLS:  true,
```
SammyOina, line 69: this should be false by default

internal/server/grpc/grpc.go
```diff
@@ -179,3 +207,68 @@ func loadX509KeyPair(certfile, keyfile string) (tls.Certificate, error) {
 	}
 	return tls.X509KeyPair(cert, key)
 }
+
+func generateCertificatesForATLS(svc *agent.Service) ([]byte, []byte, error) {
```
SammyOina, line 211: since agent has the ability to receive certs for tls, I would say we would rather parse those certs and add extensions to the certs once read

SammyOina, line 211: I don't think we need the agent service here since, I would say to pass the attestation report as bytes one generated on agent/main.go

SammyOina (changes requested):

internal/server/grpc/grpc.go
danko-miladinovic: I think we can avoid this. By adding the hash of the public key in the `report data` field of the attestation report. This is the the public key of the generated certificate (self-signed) for TLS. The idea is to have the attestation report be a part of the x.509 certificate (as an extension). Because the attestation report is signed by AMD, we can verify that the x.509 certificate is generated inside the enclave. In this way, we can avoid using server certificates from the manager and the clients can be sure that the key used in TLS communication is created in the SVM.

In order to verify that the SVM is used for a certain computation, we can hash the computation and have it be a part of the attestation report by using the `host data` field of the attestation report. This hash is supplied to the qemu command when we boot the SVM (there is a parameter `-host_data` in the qemu). This way the Agent can verify the computation (manifest) that it is getting from the manager through vsock. 

To verify that the SVM is running on the certain backend, we can use the `block ID` structure that is fed to qemu and also a part of the attestation report of SVM. This structure contatins hash of the key of the owner that started the SVM. This key can be a key of the Manager.

danko-miladinovic: The problem is that we need to insert the hash of the public key of the x.509 certificate used for TLS inside the attestation report.... Maybe we can also have the method for fetching the attestation report inside the server. How does this sound? I know that we will then have duplicate code for AR, but it seem to me to be naturan this way.

SammyOina: ok, although I suspect on the client/cli we'll need to skip verify since the self signed certs will have an unknown CA

SammyOina: I just mean to call the attestation method in main.go once service is initialized and pass the output(attestation report []byte) to this function rather than passing the while service

cmd/agent/main.go
danko-miladinovic: Changed

internal/server/grpc/grpc.go
danko-miladinovic: Yes, but the certificate is checked if it is self-signed.

pkg/clients/grpc/connect.go
```diff
@@ -29,12 +37,21 @@ var (
 	errGrpcClose   = errors.New("failed to close grpc connection")
 )
 
+var (
+	customSEVSNPExtensionOID = asn1.ObjectIdentifier{1, 2, 3, 4, 5, 6}
+	computation              = agent.Computation{}
```
SammyOina, line 42: on this I think you just need a struct with SNPPolicy and RootOFTrust. So rather than modifying agent.Computation create a new struct with the required fields

dborovcanin: customSEVSNPExtensionID defines the extension ID, so it's OK to keep it here (ideally it would be constant). As for Computation that's used here only to load policy, please create a new struct.

danko-miladinovic: Ok, I will change this. Thanks.

danko-miladinovic: Changed.

internal/server/server.go
```diff
@@ -17,12 +17,13 @@ type Server interface {
...
-	KeyFile      string `env:"SERVER_KEY"      envDefault:""`
-	ServerCAFile string `env:"SERVER_CA_CERTS" envDefault:""`
-	ClientCAFile string `env:"CLIENT_CA_CERTS" envDefault:""`
+	Host         string `env:"HOST"               envDefault:""`
+	Port         string `env:"PORT"               envDefault:""`
+	CertFile     string `env:"SERVER_CERT"        envDefault:""`
+	KeyFile      string `env:"SERVER_KEY"         envDefault:""`
+	ServerCAFile string `env:"SERVER_CA_CERTS"    envDefault:""`
+	ClientCAFile string `env:"CLIENT_CA_CERTS"    envDefault:""`
+	AttestedTLS  bool   `env:"ATTESTED_TLS"       envDefault:"false"`
```
SammyOina, line 26: agent configuration is set from manager please modify this https://github.com/ultravioletrs/cocos/blob/b6f116474d7f69f843faad1c4c060d6a5ab44e73/cmd/agent/main.go#L62-L68

SammyOina (changes requested):

internal/server/server.go
SammyOina: This will include updates on `test` dir as well and `manager` protocol buffer

pkg/manager/manager.pb.go
```diff
@@ -3,7 +3,7 @@
 
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.31.0
+// 	protoc-gen-go v1.33.0
```
SammyOina, line 6: update protoc gen go version on ci

SammyOina (changes requested):

pkg/manager/manager.pb.go
danko-miladinovic: Changed

test/manual/README.md
```diff
@@ -24,6 +24,10 @@ Open console on the host, and run
 export AGENT_GRPC_URL=localhost:7002
 export MANAGER_GRPC_URL=localhost:7001
 
+# For attested TLS also define 
+export AGENT_GRPC_MANIFEST=./test/manual/computation/computation.json # change the path to the
```
SammyOina, line 28: the comment looks incomplete

pkg/clients/grpc/connect.go
```diff
@@ -140,3 +175,99 @@ func connect(cfg Config) (*grpc.ClientConn, security, error) {
...
+		decoder := json.NewDecoder(manifest)
+		err = decoder.Decode(&attestationConfiguration)
+		if err != nil {
+			return fmt.Errorf("manifest file is malformed %w", err)
+		}
+
+		return nil
+	}
+
+	return fmt.Errorf("manifest does not exist")
```
SammyOina, line 196: define error as a variable

pkg/clients/grpc/connect.go
```diff
@@ -140,3 +175,99 @@ func connect(cfg Config) (*grpc.ClientConn, security, error) {
 	}
 	return conn, secure, nil
 }
+
+func readManifest(cfg Config) error {
+	if cfg.Manifest != "" {
+		manifest, err := os.Open(cfg.Manifest)
+		if err != nil {
+			return fmt.Errorf("failed to open manifest %w", err)
```
SammyOina, line 183: define error and use errors.Wrap

pkg/clients/grpc/connect.go
```diff
@@ -140,3 +175,99 @@ func connect(cfg Config) (*grpc.ClientConn, security, error) {
...
+		return nil
+	}
+
+	return fmt.Errorf("manifest does not exist")
+}
+
+func verifyAttestationReportTLS(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
+	cert, err := x509.ParseCertificate(rawCerts[0])
+	if err != nil {
+		return fmt.Errorf("failed to parse certficate %w", err)
```
SammyOina, line 202: same here as in previous comments

internal/server/grpc/grpc.go
```diff
@@ -179,3 +208,68 @@ func loadX509KeyPair(certfile, keyfile string) (tls.Certificate, error) {
...
+	certTemplate := &x509.Certificate{
+		SerialNumber: big.NewInt(202403311),
+		Subject: pkix.Name{
+			Organization:  []string{"Ultraviolet"},
+			Country:       []string{"Serbia"},
+			Province:      []string{""},
+			Locality:      []string{"Belgrade"},
+			StreetAddress: []string{"Bulevar Arsenija Carnojevica 103"},
+			PostalCode:    []string{"11000"},
+		},
```
SammyOina, line 239: define these values as constants

internal/server/grpc/grpc.go
```diff
@@ -179,3 +208,68 @@ func loadX509KeyPair(certfile, keyfile string) (tls.Certificate, error) {
...
+		NotBefore:             time.Now(),
+		NotAfter:              time.Now().AddDate(1, 0, 0),
```
SammyOina, line 241: define in a constant

test/manual/README.md
danko-miladinovic: Sorry, I have completed the comment.

pkg/clients/grpc/connect.go
danko-miladinovic: Done

pkg/clients/grpc/connect.go
```diff
@@ -140,3 +175,99 @@ func connect(cfg Config) (*grpc.ClientConn, security, error) {
...
+		manifest, err := os.Open(cfg.Manifest)
+		if err != nil {
+			return fmt.Errorf("failed to open manifest %w", err)
+		}
+		defer manifest.Close()
+
+		decoder := json.NewDecoder(manifest)
+		err = decoder.Decode(&attestationConfiguration)
+		if err != nil {
+			return fmt.Errorf("manifest file is malformed %w", err)
```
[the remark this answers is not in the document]
danko-miladinovic: Done

pkg/clients/grpc/connect.go
danko-miladinovic: Done

internal/server/grpc/grpc.go
danko-miladinovic: Done

internal/server/grpc/grpc.go
danko-miladinovic: Done

SammyOina (approved):

drasko (approved):
LGTM

pkg/clients/grpc/connect.go
```diff
@@ -102,34 +131,47 @@ func connect(cfg Config) (*grpc.ClientConn, security, error) {
...
+				}
+				tlsConfig.RootCAs = capool
+				secure = withTLS
 			}
-			tlsConfig.Certificates = []tls.Certificate{certificate}
-			secure = withmTLS
-		}
 
-		tc = credentials.NewTLS(tlsConfig)
+			// Loading mtls certificates file
```
drasko, line 163: mTLS

manager/qemu/qemu.go
```diff
@@ -27,18 +27,20 @@ func CreateVM(ctx context.Context, cfg Config) (*exec.Cmd, error) {
...
 	qemuCfg.SevConfig.ID = fmt.Sprintf("%s-%s", qemuCfg.SevConfig.ID, id)
 
-	// Copy firmware vars file.
-	srcFile := qemuCfg.OVMFVarsConfig.File
-	dstFile := fmt.Sprintf("%s/%s-%s.fd", cfg.TmpFileLoc, firmwareVars, id)
-	err = internal.CopyFile(srcFile, dstFile)
-	if err != nil {
-		return &exec.Cmd{}, err
+	if !cfg.KernelHash {
+		// Copy firmware vars file.
```
drasko, line 31: Remove full stop at the end

drasko (changes requested):

pkg/clients/grpc/connect.go
danko-miladinovic: Done

drasko (approved):

PR #118, opened 2024-04-19
COCOS-116 -Separated networking setup from agent start up script

### LINKED ISSUE #116: Separate networking setup from agent start up script
https://github.com/ultravioletrs/cocos/blob/3a148965559df665d508da1cccbcc9b466228c1e/init/systemd/agent_start_script.sh#L14-L18

danko-miladinovic:
# What type of PR is this?

This separates the network setup from the agent run script. The problem is defined in issue #116 

## What does this do?
This PR alters the `cocos-agent.service` file in order to run `cocos_network_setup.sh` script before the Agent is started.

Resolves #116 

No, there are no tests for this issue.

No, there is no documentation that needs altering.


init/systemd/agent_start_script.sh
```diff
@@ -1,20 +1,6 @@
...
-# The parts of the name enp0s2 mean:
-#  et - ethernet card. It means this is the ethernet interface.
-#  p - means that the interface is connected to a PCI bus.
-#  0 - the interface is connected to bus 0.
-#  s2 -the interface is connected to slot 2.
-
-# The variable ETH_IFACE value must match the name configured through QEMU parameters for the network device. 
-# The bus number and slot number are configured through QEMU device parameters, parameters
-# addr (for slot number), and bus (for bus number).
 ETH_IFACE=enp0s2
```
drasko, line 3: Instead of hardcoding this to `enp0s2`, can we use some command - like `if config`, or `addr` to get the interface name here, as it has been setup previously.

drasko (changes requested):

init/systemd/agent_start_script.sh
danko-miladinovic: If we call `dhclient` without giving it a interface name, then `dhclient` will initialize all of the network interfaces. Then we can use `route | grep '^default' | grep -o '[^ ]*$'` to fetch the name of the default interface. We only have one network interface (defined by QEMU) so this one will be the default interface. 

So, `ETH_IFACE=$(route | grep '^default' | grep -o '[^ ]*$')` but we first have to initialize all of the network interfaces by calling `dhclient`.

danko-miladinovic: After `dhclient` I can probably check if there is only one interface. To make sure that there is only one interface.

danko-miladinovic: Done

hal/linux/package/agent/agent.mk
```diff
@@ -12,14 +12,15 @@ define AGENT_BUILD_CMDS
...
 	mkdir -p $(TARGET_DIR)/var/log/cocos
 	mkdir -p $(TARGET_DIR)/cocos/
 endef
 
 define AGENT_INSTALL_INIT_SYSTEMD
-	$(INSTALL) -D -m 0644 $(@D)/init/systemd/cocos-agent.service $(TARGET_DIR)/usr/lib/systemd/system/cocos-agent.service
-	$(INSTALL) -D -m 0755 $(@D)/init/systemd/agent_start_script.sh $(TARGET_DIR)/cocos/agent_start_script.sh
+	$(INSTALL) -D -m 0640 $(@D)/init/systemd/cocos-agent.service $(TARGET_DIR)/usr/lib/systemd/system/cocos-agent.service
+	$(INSTALL) -D -m 0750 $(@D)/init/systemd/agent_start_script.sh $(TARGET_DIR)/cocos/agent_start_script.sh
+	$(INSTALL) -D -m 0750 $(@D)/init/systemd/cocos_network_setup.sh $(TARGET_DIR)/cocos/cocos_network_setup.sh
```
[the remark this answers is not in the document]
danko-miladinovic: Done

hal/linux/configs/cocos_defconfig
```diff
@@ -7,6 +7,7 @@ BR2_TARGET_GENERIC_ISSUE="Welcome to Cocos"
 BR2_PACKAGE_DHCP=y
 BR2_PACKAGE_DHCP_CLIENT=y
 BR2_INIT_SYSTEMD=y
+BR2_SYSTEM_BIN_SH_BASH=y
```
danko-miladinovic, line 10: I have added `BR2_SYSTEM_BIN_SH_BASH` so that `bash` is the default script instead of BusyBox default `ash`.

SammyOina (approved):
Tested, LGTM

init/systemd/agent_start_script.sh
```diff
@@ -1,21 +1,16 @@
...
-# The bus number and slot number are configured through QEMU device parameters, parameters
-# addr (for slot number), and bus (for bus number).
-ETH_IFACE=enp0s2
+NUM_OF_IFACE=$(ip route | grep -Eo 'dev [a-z0-9]+' | awk '{ print $2 }' | sort | uniq | wc -l)
 
-ip link set dev $ETH_IFACE up
-dhclient $ETH_IFACE
-AGENT_GRPC_HOST=$(ip -4 addr show $ETH_IFACE | grep inet | awk '{print $2}' | cut -d/ -f1)
+if [ $NUM_OF_IFACE -gt $NUM_OF_PERMITED_IFACE ]; then
+    echo "more then one network interface in the VM"
```
drasko, line 8: Just capitalize this comment - `More than ...`

drasko (changes requested):

init/systemd/agent_start_script.sh
danko-miladinovic: Changed

drasko (approved):
LGTM

PR #140, opened 2024-06-05
COCOS-135 - Refactor gRPC auth interceptors and update dependencies

### LINKED ISSUE #135: Update Auth layer in Agent

As of #128, we are using digital signatures for user authentication. However, the Auth layer is tightly coupled with API and has no interfaces we can use for testing. We should move the implementation to the root and add the corresponding interface. We still should keep interceptors on the API layer, but interceptors should call Auth abstractions rather than holding the full implementation.

Described in the step above.

Must-have

### Anything else?

N/A

SammyOina:
# What type of PR is this?
this is a refactor related to #135

## What does this do?
Introduced gRPC authentication interceptors for unary and stream calls in the agent's API to enhance security. These interceptors verify signatures based on roles before handling requests, rejecting unauthorized ones. The authentication logic was refactored, adding relevant error handling.

Added mocks for authentication interfaces to facilitate testing and replaced direct errors with wrapped ones to provide additional context.

The Makefile now includes a version pin for the Mockery tool, ensuring consistent mock generation across environments.

Updated the gRPC server setup in `grpc.go` to utilize the new interceptors. This change aims to reinforce the API's access control by authenticating users according to their roles and is expected to prevent unauthorized access to sensitive endpoints.

Lastly, dependencies in go.mod and go.sum were updated, most notably the inclusion of testify for improved testing capabilities.

- Resolves #135

Yes

No


Makefile
```diff
@@ -31,3 +32,7 @@ $(PLATFORM_INFO):
 protoc:
 	protoc -I. --go_out=. --go_opt=paths=source_relative --go-grpc_out=. --go-grpc_opt=paths=source_relative agent/agent.proto
 	protoc -I. --go_out=./pkg --go_opt=paths=source_relative --go-grpc_out=./pkg --go-grpc_opt=paths=source_relative manager/manager.proto
+
+mocks:
```
drasko, line 36: We do not need mockery installation to be automated via Makefile - please keep Makefile simple. Do installation by hand, document this in the documentation. Use Makefile to trigger the actions, not to install SW.

drasko (changes requested):

drasko (approved):
LGTM

PR #162, opened 2024-06-27
COCOS-132 - Add progress bar for algo and data uploads

### LINKED ISSUE #132: Feature: Add progress bar for algo and data uploads

User needs to be able to see the percentage of file uploads on cli for visibilirt

User needs to be able to see the percentage of file uploads on cli for visibility

Must-have

### Anything else?

_No response_

danko-miladinovic:
# What type of PR is this?

This is a feature because it adds the following functionality:
- Adds a progress bar so that the user can see how much of the algorithm and data is uploaded to the Agent.
## What does this do?
This PR adds a `ProgressBar` struct that is used to draw the progress bar onto the terminal. The example of the progress bar can be seen here:
`Uploading data... 25% [=====>                        ]`

- Resolves #132 

This PR can be tested with existing manual tests.
This feature does not need to be documented, because it is a standard feature.

### DIFF
--- a/pkg/progressbar/progressbar.go
+++ b/pkg/progressbar/progressbar.go
@@ -23,8 +23,10 @@ const (
 	bufferSize      = 1024 * 1024
 )
 
-var _ streamSender = (*algoClientWrapper)(nil)
-var _ streamSender = (*dataClientWrapper)(nil)
+var (
+	_ streamSender = (*algoClientWrapper)(nil)
+	_ streamSender = (*dataClientWrapper)(nil)
+)
 
 type streamSender interface {
 	Send(interface{}) error
--- a/pkg/progressbar/progressbar.go
+++ b/pkg/progressbar/progressbar.go
@@ -1,0 +1,217 @@
+package progressbar
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+
+	"github.com/ultravioletrs/cocos/agent"
+	"golang.org/x/term"
+)
+
+const (
+	progressBarDots = "... "
+	leftBracket     = "["
+	rightBracket    = "]"
+	head            = ">"
+	body            = "="
+	bodyPadding     = "."
+	bufferSize      = 1024 * 1024
+)
+
+type ProgressBar struct {
+	numberOfBytes           int
+	currentUploadedBytes    int
+	currentUploadPercentage int
+	description             string
+	maxWidth                int
+}
+
+func New() *ProgressBar {
+	return &ProgressBar{}
+}
+
+func (p *ProgressBar) SendAlgorithm(description string, buffer *bytes.Buffer, stream *agent.AgentService_AlgoClient) error {
+	p.currentUploadedBytes = 0
+	p.currentUploadPercentage = 0
+	p.numberOfBytes = buffer.Len()
+	p.description = description
+
+	buf := make([]byte, bufferSize)
+
+	for {
+		n, err := buffer.Read(buf)
+		if err == io.EOF {
+			if _, err := io.WriteString(os.Stdout, "\n"); err != nil {
+				return err
+			}
+			break
+		}
+		if err != nil {
+			return err
+		}
+
+		if p.currentUploadedBytes < p.numberOfBytes {
+			p.currentUploadedBytes = p.currentUploadedBytes + n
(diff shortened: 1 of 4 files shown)


pkg/progressbar/progressbar.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	if _, err := (*stream).CloseAndRecv(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (p *ProgressBar) SendData(description string, buffer *bytes.Buffer, stream *agent.AgentService_DataClient) error {
```
drasko, line 80: There is a lot of repeating code in these two functions. Can you make one common function with parameter algo/data to unify this?

pkg/sdk/agent.go
```diff
@@ -90,28 +77,14 @@ func (sdk *agentSDK) Data(ctx context.Context, dataset agent.Dataset, privKey *r
...
-		}
-
-		err = stream.Send(&agent.DataRequest{Dataset: buf[:n]})
-		if err != nil {
-			return err
-		}
-	}
-
-	if _, err := stream.CloseAndRecv(); err != nil {
+	progressbar := progressbar.New()
```
drasko, line 85: Do you need to create a new one, or you can reuse/reset old one if not used anymore?

drasko (changes requested):

pkg/progressbar/progressbar.go
SammyOina: @danko-miladinovic consider something like this 
```go
type StreamSender interface {
    Send(interface{}) error
    CloseAndRecv() (interface{}, error)
}

func (p *ProgressBar) sendData(description string, buffer *bytes.Buffer, stream StreamSender, createRequest func([]byte) interface{}) error {
    p.reset(description, buffer.Len())

    buf := make([]byte, bufferSize)

    for {
        n, err := buffer.Read(buf)
        if err == io.EOF {
            if _, err := io.WriteString(os.Stdout, "\n"); err != nil {
                return err
            }
            break
        }
        if err != nil {
            return err
        }

        p.updateProgress(n)

        if err := stream.Send(createRequest(buf[:n])); err != nil {
            return err
        }

        if err := p.renderProgressBar(); err != nil {
            return err
        }
    }

    _, err := stream.CloseAndRecv()
    return err
}

func (p *ProgressBar) reset(description string, totalBytes int) {
    p.currentUploadedBytes = 0
    p.currentUploadPercentage = 0
    p.numberOfBytes = totalBytes
    p.description = description
}

func (p *ProgressBar) updateProgress(bytesRead int) {
    if p.currentUploadedBytes < p.numberOfBytes {
        p.currentUploadedBytes += bytesRead
        p.currentUploadPercentage = float64(p.currentUploadedBytes) * 100.0 / float64(p.numberOfBytes)
    }
}

func (p *ProgressBar) SendAlgorithm(description string, buffer *bytes.Buffer, stream *agent.AgentService_AlgoClient) error {
    return p.sendData(description, buffer, *stream, func(data []byte) interface{} {
        return &agent.AlgoRequest{Algorithm: data}
    })
}

func (p *ProgressBar) SendData(description string, buffer *bytes.Buffer, stream *agent.AgentService_DataClient) error {
    return p.sendData(description, buffer, *stream, func(data []byte) interface{} {
        return &agent.DataRequest{Dataset: data}
    })
}
```

drasko, line 80: Do we need to export this function?

drasko (changes requested):

pkg/sdk/agent.go
danko-miladinovic: Yes, progress bar is used in Agent SDK so we need to create a new one every time the users uses the CLI to send data and algo.

pkg/progressbar/progressbar.go
```diff
@@ -0,0 +1,232 @@
...
+const (
+	progressBarDots = "... "
+	leftBracket     = "["
+	rightBracket    = "]"
+	head            = ">"
+	body            = "="
+	bodyPadding     = "."
+	bufferSize      = 1024 * 1024
+)
+
```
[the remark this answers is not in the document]
danko-miladinovic: Done

dborovcanin (approved):

PR #175, opened 2024-07-10
COCOS-168 - Allow Computations With No Datasets To Run

### LINKED ISSUE #168: Feature: Allow state-machine transition from algo-received state to running state incase of no dataset provided

Agent will be stuck on receiving dataset even when no dataset is provided in manifest

Agent should be able to transition to running state in case of no algo is provided. Here are some samples to start on

```go
sm.Transitions[receivingAlgorithm][dataReceived] = running
```

```go
if dataFileExists {
    sm.SendEvent(algorithmReceived)
} else {
    sm.SendEvent(dataReceived)
}
```

Must-have

### Anything else?

_No response_

rodneyosodo:
# What type of PR is this?

This is a feature because it adds running computation when dataset is empty

## What does this do?

The updates enable state-machine transitions from the `receivingAlgorithm` state to the `running` state if no dataset is provided.

- Resolves #168 

Yes, I have included tests for my changes.

No need for documentation since this is an expected scenario

### Notes

- **Refactor**
  - Updated state machine transitions to improve control flow.

- **Bug Fixes**
  - Changed types from `int` to `uint8` for state and event to prevent possible overflows.

- **Tests**
  - Modified tests to reflect new state machine transitions and event handling.
  
```mermaid
sequenceDiagram
    participant User
    participant Agent
    participant StateMachine
    User->>Agent: Provide Algorithm
    Note right of Agent: Checking dataset length
    alt Dataset length is 0
        Agent->>StateMachine: Send algoReceivedNoData event
        StateMachine->>StateMachine: Transition to running state
    else Dataset length > 0
        Agent->>StateMachine: Send algorithmReceived event
        StateMachine->>StateMachine: Transition to receivingAlgorithm state
    end
    StateMachine->>User: Acknowledge state transition
```

### DIFF
--- a/agent/service.go
+++ b/agent/service.go
@@ -87,12 +87,6 @@ func New(ctx context.Context, logger *slog.Logger, eventSvc events.Service, cmp 
 	svc.sm.StateFunctions[complete] = svc.publishEvent("in-progress", json.RawMessage{})
 	svc.sm.StateFunctions[running] = svc.runComputation
 
-	if len(cmp.Datasets) == 0 {
-		svc.sm.StateFunctions[receivingData] = func() {
-			svc.sm.SendEvent(dataReceived)
-		}
-	}
-
 	svc.computation = cmp
 	svc.sm.SendEvent(manifestReceived)
 	return svc
@@ -130,6 +124,11 @@ func (as *agentService) Algo(ctx context.Context, algorithm Algorithm) error {
 	}
 
 	as.algorithm = f.Name()
+
+	if len(as.computation.Datasets) == 0 {
+		as.sm.SendEvent(algoReceivedNoData)
+		return nil
+	}
 
 	if as.algorithm != "" {
 		as.sm.SendEvent(algorithmReceived)
--- a/agent/service.go
+++ b/agent/service.go
@@ -125,6 +125,10 @@ func (as *agentService) Algo(ctx context.Context, algorithm Algorithm) error {
 
 	as.algorithm = f.Name()
 
+	if len(as.computation.Datasets) == 0 {
+		as.sm.SendEvent(dataReceived)
+	}
+
 	if as.algorithm != "" {
 		as.sm.SendEvent(algorithmReceived)
 	}
--- a/agent/service.go
+++ b/agent/service.go
@@ -87,6 +87,12 @@ func New(ctx context.Context, logger *slog.Logger, eventSvc events.Service, cmp 
 	svc.sm.StateFunctions[complete] = svc.publishEvent("in-progress", json.RawMessage{})
 	svc.sm.StateFunctions[running] = svc.runComputation
 
+	if len(cmp.Datasets) == 0 {
+		svc.sm.StateFunctions[receivingData] = func() {
+			svc.sm.SendEvent(dataReceived)
+		}
+	}
+
 	svc.computation = cmp
 	svc.sm.SendEvent(manifestReceived)
 	return svc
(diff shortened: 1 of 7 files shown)


SammyOina (approved):

test/manual/algo/addition.py
SammyOina, line 1: add a readme file here

SammyOina (changes requested):

test/manual/algo/addition.py
rodneyosodo: There is a [README](https://github.com/ultravioletrs/cocos/blob/main/test/manual/algo/README.md) file aready

agent/service.go
```diff
@@ -126,6 +126,10 @@ func (as *agentService) Algo(ctx context.Context, algorithm Algorithm) error {
 
 	as.algorithm = f.Name()
 
+	if len(as.computation.Datasets) == 0 {
+		as.sm.SendEvent(dataReceived)
```
drasko, line 130: This is a hack. I would rather implement more correct approach, and that is that state machine goes directly to running if Dataset is 0, i.e. to go in data receiving stage only if there are some data to be received - otherwise go to running mode.

drasko (changes requested):

agent/service.go
rodneyosodo: Yes. It goes to `running` state. This is the event that triggers the change of state from `receivingAlgorithm` to `running`

SammyOina: you could move this check into the state call back functions for  algo, such that it is declares in `New` agent function which has the manifest

SammyOina (changes requested):

agent/service.go
```diff
@@ -88,6 +88,12 @@ func New(ctx context.Context, logger *slog.Logger, eventSvc events.Service, cmp
 	svc.sm.StateFunctions[complete] = svc.publishEvent("in-progress", json.RawMessage{})
 	svc.sm.StateFunctions[running] = svc.runComputation
 
+	if len(cmp.Datasets) == 0 {
+		svc.sm.StateFunctions[receivingData] = func() {
+			svc.sm.SendEvent(dataReceived)
```
drasko, line 93: Why are you sending `dataReceived` even when there was no data? This is a hack, and I wrote this before.

I do not like this - we should rather change the state machine not to even go in `receivingData` if there is no data, but to directly switch to `receivingAlgo` state

agent/state.go
```diff
@@ -63,6 +63,7 @@ func NewStateMachine(logger *slog.Logger) *StateMachine {
 
 	sm.Transitions[receivingAlgorithm] = make(map[event]state)
 	sm.Transitions[receivingAlgorithm][algorithmReceived] = receivingData
```
drasko, line 65: Do not switch to `receivingData` state if there is no data`

drasko (changes requested):

agent/service.go
```diff
@@ -126,6 +126,11 @@ func (as *agentService) Algo(ctx context.Context, algorithm Algorithm) error {
 
 	as.algorithm = f.Name()
 
+	if len(as.computation.Datasets) == 0 {
+		as.sm.SendEvent(algoReceivedNoData)
```
drasko, line 130: I do not think we need this

agent/state.go
```diff
@@ -63,6 +64,7 @@ func NewStateMachine(logger *slog.Logger) *StateMachine {
...
+	sm.Transitions[receivingAlgorithm][algoReceivedNoData] = running
```
drasko, line 67: Put an `if` here and transition to data receiving or running state im dépendance of data length

drasko (changes requested):

agent/state.go
SammyOina: @rodneyosodo https://github.com/ultravioletrs/cocos/blob/d76074ae41b03c794f0dcb3da73a4ec0f67906ff/agent/state.go#L48-L77 change this to accept manifest, and define state transitions based on contents of manifest and as a result there will be no need to change service

SammyOina (approved):

dborovcanin (approved):

PR #179, opened 2024-07-18
ALGO STDOUT-STDERR

SammyOina:
# What type of PR is this?

- Related Issue #
- Resolves #

### DIFF
--- a/agent/algorithm/binary/binary.go
+++ b/agent/algorithm/binary/binary.go
@@ -12,7 +12,6 @@ import (
 
 	"github.com/ultravioletrs/cocos/agent/algorithm"
 	"github.com/ultravioletrs/cocos/agent/events"
-	"github.com/ultravioletrs/cocos/pkg/socket"
 )
 
 const socketPath = "unix_socket"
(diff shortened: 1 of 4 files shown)


dborovcanin:
Let's evaluate this approach:
- simplifies the flow since we rely on "standard" std{in, our, err} approach
- enables unified approach across different runners: binary, Docker, WASM runtime
- less invasive on algorithm - we do not expect algo to be written in such a way that socket is open for streaming process info

However, there are some drawbacks to this approach:
- limited output to a single stream (err for logs and out for result) provides less insight into the algorithm execution
- it still requires code to be written in a certain way, i.e., to redirect all logs to stdout rather than a file

Having this in mind, it's worth exploring if there is a way to "fake" the filesystem for those algorithms that require files for logging. @SammyOina Can you check it? Also, try and test how this approach works with Docker.

dborovcanin:
Finally, I think it would be an acceptable requirement (though definitely not ideal, ofc) for the algo provider to write all the logs to the stderr, since we do not support a file system.

PR #187, opened 2024-07-29
PRISM-312 : Fetch Backend Information

WashingtonKK:
To be merged with: https://github.com/ultravioletrs/prism/pull/398


manager/service.go
```diff
@@ -44,6 +46,39 @@ var (
 	ErrFailedToCalculateHash = errors.New("error while calculating the hash of the computation")
 )
 
+type Vmpl struct {
+	Value uint32 `json:"value"`
+}
+
+type SnpPolicy struct {
```
SammyOina, line 53: why not use https://github.com/google/go-sev-guest/blob/00148f763a981a74331226d462983bd9e9b903c4/proto/check/check.pb.go#L44 since it is already defined

manager/service.go
```diff
@@ -44,6 +46,39 @@ var (
...
+	HostData                  []byte `json:"host_data"`
+	ReportIDMA                []byte `json:"report_id_ma"`
+	ChipID                    []byte `json:"chip_id"`
+	MinimumBuild              uint32 `json:"minimum_build"`
+	MinimumVersion            string `json:"minimum_version"`
+	PermitProvisionalFirmware bool   `json:"permit_provisional_firmware"`
+	RequireIDBlock            bool   `json:"require_id_block"`
+}
+
+type RootOfTrust struct {
```
SammyOina, line 71: same here

SammyOina (changes requested):

manager/service.go
```diff
@@ -160,6 +164,22 @@ func (ms *managerService) Stop(ctx context.Context, computationID string) error
 	return nil
 }
 
+func (ms *managerService) FetchBackendInfo() ([]byte, error) {
+	cmd := exec.Command("sudo", "../../scripts/backend_info/backend_info", "--policy", "1966081")
```
SammyOina, line 168: should be an env variable, and the default should be in the build dir as other bins. Also sudo is not required

scripts/backend_info/backend_info.json
SammyOina, line 1: also remove the orignal `/test/computation` file

SammyOina (changes requested):

manager/service.go
```diff
@@ -160,6 +164,22 @@ func (ms *managerService) Stop(ctx context.Context, computationID string) error
...
+
+	_, err := cmd.Output()
+	if err != nil {
+		return nil, err
+	}
+
+	f, err := os.ReadFile("./scripts/backend_info/backend_info.json")
```
SammyOina, line 175: path should be relative to location of the backend_info bin

SammyOina (changes requested):

SammyOina (changes requested):
also update test/computation to display backend info

SammyOina (approved):

SammyOina (approved):
please update or add relevant docs on cocos docs

drasko (approved):
LGTM

PR #191, opened 2024-08-07
PRISM-337: Add Filename to Dataset

WashingtonKK:
# What type of PR is this?

This is a feature because it adds filename to dataset.

## What does this do?

This adds filename to dataset, to allow the filename to be passed from backend to manager.

- Related Issue #

No need for tests


SammyOina (approved):

drasko (approved):
LGTM

PR #197, opened 2024-08-09
COCOS-181 - Add failed state for algo execution

### LINKED ISSUE #181: Feature: Transition to Different State on Failed Computation Run

When computation run fails, agent produces an event indicating failed run, however, the next state after this is `resultsReady` which may mis-inform the user.

image

This will enhance the user experience and allow the user to know that the run failed, and no results are available.

Should-have

### Anything else?

_No response_

SammyOina:
# What type of PR is this?
This a feature as it adds a new failed state to agent

## What does this do?
- **New Features**
	- Enhanced error handling with the introduction of a new error for consumed results.
	- Added support for failure states in the computation process.

- **Improvements**
	- Refined state machine logic to handle computation failures and transitions more effectively.
	- Improved clarity of comments and descriptions for better understanding of service behavior.

- Resolves #181 

No

Will update cocos docs

### DIFF
--- a/agent/service.go
+++ b/agent/service.go
@@ -51,7 +51,7 @@ var (
 	// ErrFileNameMismatch provided dataset filename does not match filename in manifest.
 	ErrFileNameMismatch = errors.New("malformed data, filename does not match manifest")
 	// ErrAllResultsConsumed indicates all results have been consumed.
-	ErrAllResultsConsumed = errors.New("all results have been consumed bu declared consumers")
+	ErrAllResultsConsumed = errors.New("all results have been consumed by declared consumers")
 )
 
 // Service specifies an API that must be fullfiled by the domain service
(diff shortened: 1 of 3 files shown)


WashingtonKK (commented):
@SammyOina 
What do you think about this:

When a user attempts to fetch results before they are ready, return a warning that results are not ready and do not transition to the next state on the state machine.

Can this also be handled in this PR?

SammyOina:
> @SammyOina What do you think about this:
> 
> When a user attempts to fetch results before they are ready, return a warning that results are not ready and do not transition to the next state on the state machine.
> 
> Can this also be handled in this PR?

agent already returns results not ready warning. agent will not transition from running state unless results are ready or algo execution failed.

WashingtonKK (changes requested):
image

We have two indications of failed, I would say we only have:

`Running -> Failed`

And remove: 

`Failed -> failed`

SammyOina:
> image
> 
> We have two indications of failed, I would say we only have:
> 
> `Running -> Failed`
> 
> And remove:
> 
> `Failed -> failed`

Failed state exits for two reasons, one is for result consumers to see the reason for failed computation, another is to allow agent to remain in this state after running, it cannot remain in failed state perpetually. i.e running failed is not an actual state. agent state machine is still running. see https://github.com/SammyOina/cocos-ai/blob/cocos-181/agent/state.go

PR #222, opened 2024-08-28
NOISSUE - Cache and retry message sending

SammyOina:
# What type of PR is this?
This is a feature which adds retry for logs and events

## What does this do?
- **New Features**
	- Enhanced connection management with the introduction of a `SafeConn` structure for improved safety and robustness.
	- Added a message retry mechanism to ensure reliable message delivery in concurrent environments.
	- Introduced caching of messages that fail to send immediately, allowing for retries.

- **Bug Fixes**
	- Simplified the `Close` method to improve error handling and method contract consistency.

- **Documentation**
	- Updated function signatures to reflect changes in connection handling and service initialization.

- Related Issue #214


dborovcanin (changes requested):

internal/logger/protohandler.go
```diff
@@ -72,27 +85,63 @@ func (h *handler) Handle(_ context.Context, r slog.Record) error {
...
 		b, err := proto.Marshal(&agentLog)
 		if err != nil {
 			return err
 		}
 
-		// Write the chunk to the writer
-		if _, err := h.w.Write(b); err != nil {
-			return err
+		h.mutex.Lock()
+		h.w.Mu.Lock()
```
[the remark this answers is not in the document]
SammyOina: it is concurrent safe my mistake

agent/events/events.go
```diff
@@ -58,12 +66,45 @@ func (s *service) SendEvent(event, status string, details json.RawMessage) error
...
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
+
+	for i := 0; i < len(s.cachedMessages); {
+		if _, err := s.conn.Write(s.cachedMessages[i]); err != nil {
+			i++
+		} else {
+			s.cachedMessages = append(s.cachedMessages[:i], s.cachedMessages[i+1:]...)
+		}
+	}
```
dborovcanin, line 105: Let's use a tmp map for this.

dborovcanin (changes requested):

agent/events/events.go
dborovcanin: No need, we can simplify with something like this:
```go

func (s *service) retrySendCachedMessages() {
	s.mutex.Lock()
	defer s.mutex.Unlock()
	tmp := [][]byte{}
	for _, msg := range s.cachedMessages {
		if _, err := s.conn.Write(msg); err != nil {
			tmp = append(tmp, msg)
		}
	}
	s.cachedMessages = tmp
}
```

drasko:
@dborovcanin please follow this one and let's move it forward ASAP

dborovcanin (approved):

internal/logger/protohandler.go
```diff
@@ -72,27 +80,59 @@ func (h *handler) Handle(_ context.Context, r slog.Record) error {
...
+	defer h.mutex.Unlock()
+
+	for i := 0; i < len(h.cachedMessages); {
+		_, err := h.w.Write(h.cachedMessages[i])
+		if err != nil {
+			i++
+		} else {
+			h.cachedMessages = append(h.cachedMessages[:i], h.cachedMessages[i+1:]...)
+		}
+	}
```
dborovcanin, line 124: Why did you revert this from the discussed approach?

dborovcanin (changes requested):

dborovcanin (approved):

PR #230, opened 2024-09-04
NOISSUE - Stop computation on complete state

SammyOina:
# What type of PR is this?
This is a feature that stops vm when computation is complete

## What does this do?
stops computation on complete state


drasko:
We will not do this

PR #252, opened 2024-09-23
NOISSUE - Reduce message loss via vsock with acks

SammyOina:
# What type of PR is this?

## What does this do?
- **New Features**
  - Introduced a new connection handling mechanism with acknowledgment support for improved message reliability.
  - Enhanced event handling with updated timing for event dispatching.

- **Bug Fixes**
  - Improved error handling during connection reads and logging of broken connections.

- **Refactor**
  - Streamlined connection handling logic and updated method names for clarity.
  - Updated state transition logic for better state management.

- **Documentation**
  - Improved logging statements for better readability and debugging.

### DIFF
--- a/internal/vsock/client.go
+++ b/internal/vsock/client.go
@@ -4,51 +4,58 @@
 package vsock
 
 import (
-	"bufio"
+	"encoding/binary"
 	"fmt"
 	"io"
 	"log"
-	"strings"
+	"net"
 	"sync"
 	"time"
 
-	"github.com/mdlayher/vsock"
+	"google.golang.org/protobuf/proto"
 )
 
-var _ io.Writer = (*AckWriter)(nil)
-
 const (
-	maxRetries = 3
-	retryDelay = time.Second
+	maxRetries     = 3
+	retryDelay     = time.Second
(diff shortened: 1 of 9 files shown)


internal/vsock/client.go
```diff
@@ -0,0 +1,230 @@
...
+		return 0, fmt.Errorf("message size exceeds maximum allowed size of %d bytes", maxMessageSize)
+	}
+
+	aw.ackMu.Lock()
+	messageID := aw.nextID
+	aw.nextID++
+	aw.ackMu.Unlock()
+
+	ackCh := make(chan bool, 1)
+	aw.ackMu.Lock()
```
dborovcanin, line 74: Don't unlock. It's more expensive than the allocation for which you are releasing it, it's not worth it.

dborovcanin (changes requested):
This state machine is becoming too complex, with too many mutex ops and supporting code (such as vsock client in this PR). We also need tests, and it will be difficult since we use structs all over the place. We also need to improve encapsulation and separate responsibilities. Please open a ticket for refactoring Agent's state machine, we won't do it in this PR.

dborovcanin (approved):

PR #287, opened 2024-10-17
NOISSUE - Refactor manager events and detangle service

SammyOina:
# What type of PR is this?
This is a refactor on manager service 

## What does this do?
Abstracts events and detangles  manager dependencies

- **New Features**
	- Introduced a new structure for event logging with `EventsLogs`.
	- Added a new command for generating Go code from `events.proto`.
	- Implemented a method to report broken connections.

- **Bug Fixes**
	- Updated logging mechanisms across various algorithm packages for improved clarity.

- **Refactor**
	- Restructured the `manager` package, including changes to method signatures and import paths.

- **Documentation**
	- Updated test files to reflect changes in structures and methods.

### DIFF
--- a/agent/service.go
+++ b/agent/service.go
@@ -180,11 +180,11 @@ func New(ctx context.Context, logger *slog.Logger, eventSvc events.Service, cmp 
 }
 
 func (as *agentService) Algo(ctx context.Context, algo Algorithm) error {
-	as.mu.Lock()
-	defer as.mu.Unlock()
 	if as.sm.GetState() != ReceivingAlgorithm {
 		return ErrStateNotReady
 	}
+	as.mu.Lock()
+	defer as.mu.Unlock()
 	if as.algorithm != nil {
 		return ErrAllManifestItemsReceived
 	}
@@ -263,11 +263,11 @@ func (as *agentService) Algo(ctx context.Context, algo Algorithm) error {
 }
 
 func (as *agentService) Data(ctx context.Context, dataset Dataset) error {
-	as.mu.Lock()
-	defer as.mu.Unlock()
 	if as.sm.GetState() != ReceivingData {
 		return ErrStateNotReady
 	}
+	as.mu.Lock()
+	defer as.mu.Unlock()
 	if len(as.computation.Datasets) == 0 {
 		return ErrAllManifestItemsReceived
 	}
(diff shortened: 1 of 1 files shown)


manager/vm/vm.go
```diff
@@ -17,10 +18,40 @@ type VM interface {
...
 	GetConfig() interface{}
 }
 
 //go:generate mockery --name Provider --output=./mocks --filename provider.go --quiet --note "Copyright (c) Ultraviolet \n // SPDX-License-Identifier: Apache-2.0"
-type Provider func(config interface{}, logsChan chan *manager.ClientStreamMessage, computationId string) VM
+type Provider func(config interface{}, eventSender EventSender, computationId string) VM
+
+type EventsLogs interface {
+	IsEventLog() bool
+}
```
dborovcanin, line 31: This interface seems weird. What's the purpose of this method except to provide a wrapper around events and logs? Do events and logs share any common behavioral pattern (say, can be seralized)?

dborovcanin (changes requested):

manager/events/vsock.go
```diff
@@ -0,0 +1,121 @@
...
+		lis:                    l,
+		reportBrokenConnection: reportBrokenConnection,
+		logger:                 logger,
+		eventsChan:             eventsChan,
+	}
+}
+
+func (e *events) Listen() {
+	for {
+		conn, err := e.lis.Accept()
```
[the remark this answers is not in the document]
SammyOina: it handles multiple clients so it should not return until the service ends

manager/vm/vm.go
SammyOina: Yes they can be serilized, but there is is not higher purpose to this interface, maybe removing the type here is the way to go so its just interface{}

agent/service.go
```diff
@@ -178,6 +180,8 @@ func New(ctx context.Context, logger *slog.Logger, eventSvc events.Service, cmp
 }
 
 func (as *agentService) Algo(ctx context.Context, algo Algorithm) error {
+	as.mu.Lock()
+	defer as.mu.Unlock()
```
dborovcanin, line 184: Can you move this lock closer to where it is used, i.e. wehere you access (and set) algorithm.

dborovcanin (changes requested):

dborovcanin (approved):

PR #363, opened 2025-01-29
COCOS-363 - Add verification of vTPM attestation to CLI

### LINKED ISSUE #352: Feature: Add CLI option for verifying the vTPM attestation.

The CVM will now launch with a vTPM. The CLI now must be capable of verifying the vTPM attestation report.

The CLI needs to be extended with an option to verify the vTPM attestation. The [go-tpm-tools](https://github.com/google/go-tpm-tools) library can be used.

Must-have

### Anything else?

_No response_

dorcaslitunya:
# What type of PR is this?

## What does this do?
CLI attestation changes

Issue #352 

- Related Issue #352 
- Resolves #352 

No, draft PR.

Yes


cli/attestation.go
```diff
@@ -473,19 +562,347 @@ func (cli *CLI) NewValidateAttestationValidationCmd() *cobra.Command {
...
+}
+
+func sevsnpverify(cmd *cobra.Command, args []string) error {
+
+	cmd.Println("Checking attestation")
+
+	attestationFile = string(args[0])
+
+	if err := parseConfig(); err != nil {
+		printError(cmd, "Error parsing config: %v ❌ ", err)
```
SammyOina, line 575: print the error on the caller function

cli/attestation.go
```diff
@@ -119,8 +129,23 @@ var (
...
+	output              string
+	input               string
+	nonce               []byte
+	format              string
+	teeNonce            []byte
 )
 
+type alwaysError struct {
+	err error
+}
```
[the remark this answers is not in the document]
dorcaslitunya: This is because, some functions that use it(dataInput and dataOutput) need something that has an interface of an Reader. The Reader interface needs expects an implementation with a Read() method attached to it. In the case of an unsuccessful read/write, it cannot return an error in its basic form since it does not mimic the Reader interface which is done by the read() and write() function tied to the struct.

 So in the case of an error, we need a struct that can implement a Read()/Write() method so as to appropriately return an error when reading fails that follows with the Reader interface.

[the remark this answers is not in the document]
SammyOina: is the datainput and output from actual files or why do you need to implement the interface. if no read is actually happening we can just use io.Discard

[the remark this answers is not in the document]
dorcaslitunya: The data input and output is from actual files. I need to implement the interface when the read fails, since the two functions(dataInput() and dataOutput()) return a type of io.Reader..so whatever happens(whether we successfully read or fail to read) we need to return sth in the fashion of a Reader

go.mod
```diff
@@ -8,7 +8,8 @@ require (
 	github.com/fatih/color v1.18.0
 	github.com/go-kit/kit v0.13.0
 	github.com/gofrs/uuid v4.4.0+incompatible
-	github.com/google/go-sev-guest v0.12.1
+	github.com/google/go-sev-guest v0.11.1
```
SammyOina, line 11: use the later version of go-sev-guest since this is a downgrade

SammyOina (changes requested):

cli/attestation.go
```diff
@@ -473,17 +552,195 @@ func (cli *CLI) NewValidateAttestationValidationCmd() *cobra.Command {
...
+		return err
+	}
+	cryptoPub, err := pub.Key()
+	if err != nil {
+		return err
+	}
+
+	var validateOpts interface{}
+	switch attestation.GetTeeAttestation().(type) {
+	case *tpmAttest.Attestation_TdxAttestation:
```
danko-miladinovic, line 625: Can we remove the TDX attestation verification option? We will probably add it in the future, but for now I think it should be removed.

danko-miladinovic (changes requested):

go.mod
dorcaslitunya: The later version does not have support fr some of our dependencies. Rolled back to v0.11.1

SammyOina (changes requested):
image


help menu is shown when validation fails, also error message appears twice

cli/attestation.go
```diff
@@ -473,17 +554,183 @@ func (cli *CLI) NewValidateAttestationValidationCmd() *cobra.Command {
...
+		return fmt.Errorf("error parsing hashes: %v ❌ ", err)
+	}
+	if err := parseFiles(); err != nil {
+		return fmt.Errorf("error parsing files: %v ❌ ", err)
+	}
+	// This format is the attestation report in AMD's specified ABI format, immediately
+	// followed by the certificate table bytes.
+	if len(attestation) < abi.ReportSize {
+		msg := color.New(color.FgRed).Sprintf("attestation contents too small (0x%x bytes). Want at least 0x%x bytes ❌ ", len(attestation), abi.ReportSize)
+		fmt.Println(msg) // Directly printing the error message
```
SammyOina, line 578: print error at the caller function, and use cmd.Println as in other cases

cli/attestation.go
```diff
@@ -473,17 +554,183 @@ func (cli *CLI) NewValidateAttestationValidationCmd() *cobra.Command {
...
+	if output == "" {
+		return os.Stdout
+	}
+
+	file, err := os.Create(output)
+	if err != nil {
+		return alwaysError{err}
+	}
+	return file
 }
```
SammyOina, line 734:
```suggestion
func openInputFile() (io.Reader, error) {
    if tmpAttestationFile == "" {
        return nil, errEmptyFile
    }
    return os.Open(tmpAttestationFile)
}

func createOutputFile() (io.Writer, error) {
    if output == "" {
        return os.Stdout, nil
    }
    return os.Create(output)
}
```

cli/attestation.go
```diff
@@ -473,17 +554,183 @@ func (cli *CLI) NewValidateAttestationValidationCmd() *cobra.Command {
...
-		printError(cmd, "Failed to mark flag as required: %v ❌ ", err)
+	return nil
+}
+
+func vtpmverify(args []string) error {
+	tpmAttestationFile = string(args[0])
+	attestationBytes, err := io.ReadAll(dataInput())
+	if err != nil {
+		return err
+	}
```
SammyOina, line 663:
```suggestion
    input, err := openInputFile()
    if err != nil {
        return err
    }
    if closer, ok := input.(*os.File); ok {
        defer closer.Close()
    }
	attestationBytes, err := io.ReadAll(input)
	if err != nil {
		return err
	}
```

cli/attestation.go
```diff
@@ -473,17 +554,183 @@ func (cli *CLI) NewValidateAttestationValidationCmd() *cobra.Command {
...
+	if err != nil {
+		return fmt.Errorf("verifying attestation: %w", err)
+	}
+	out, err := marshalOptions.Marshal(ms)
+	if err != nil {
 		return nil
 	}
+	if _, err := dataOutput().Write(out); err != nil {
+		return fmt.Errorf("failed to write verified attestation report: %v", err)
+	}
```
SammyOina, line 696:
```suggestion
    output, err := createOutputFile()
    if err != nil {
        return err
    }
    if closer, ok := output.(*os.File); ok {
        defer closer.Close()
    }
	if _, err := output.Write(out); err != nil {
		return fmt.Errorf("failed to write verified attestation report: %v", err)
	}
```

SammyOina: this will remove need for the alwaysError struct which does not seem idiomatic in go

SammyOina (changes requested):

SammyOina (approved):

cli/attestation.go
```diff
@@ -473,17 +552,184 @@ func (cli *CLI) NewValidateAttestationValidationCmd() *cobra.Command {
...
+	if err != nil {
+		return err
+	}
+	attestation := &tpmAttest.Attestation{}
 
-	if err := cmd.MarkFlagRequired("product"); err != nil {
-		printError(cmd, "Failed to mark flag as required: %v ❌ ", err)
+	if format == "binarypb" {
+		err = proto.Unmarshal(attestationBytes, attestation)
+	} else if format == "textproto" {
```
SammyOina, line 685: use string constants for binarypb and textproto. also fix CI

SammyOina (changes requested):

SammyOina (approved):

dborovcanin (approved):

PR #372, opened 2025-02-11
COCOS-369 - Enable compiling gRPC clients without cGo

### LINKED ISSUE #369: Move aTLS to a separate module
Importing CoCoS Agent results in many cases with CGO dependency propagation due to aTLS implementation. This can be avoided if we have aTLS lib as a separate module. Let's try to keep it within the CoCoS repository, making CoCoS multimodule repo.

SammyOina:
# What type of PR is this?

## What does this do?
- **Chores**
  - Updated runtime environment and dependency configurations to boost overall stability and compatibility.

- Resolves #369


dborovcanin:
Please resolve conflicts.

pkg/clients/grpc/grpc.go
```diff
@@ -239,52 +215,3 @@ func ReadAttestationPolicy(manifestPath string, attestationConfiguration *check.
 
 	return ErrAttestationPolicyMissing
 }
-
-func CustomDialer(ctx context.Context, addr string) (net.Conn, error) {
```
dborovcanin, line 243: Is this not needed anymore, i.e. it was not necessary in the first place?

pkg/clients/grpc/atls_stub.go
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Ultraviolet
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build !cgo
```
dborovcanin, line 4: Good idea. :)

pkg/clients/grpc/grpc.go
SammyOina: we need the custom dialer since we add tls extensions and standard go library does not allow for this

dborovcanin (approved):

PR #417, opened 2025-04-02
NOISSUE - Reset agent statemachine when computation is stopped

SammyOina:
# What type of PR is this?

- Related Issue #
- Resolves #


agent/statemachine/state.go
```diff
@@ -89,6 +90,20 @@ func (sm *stateMachine) Start(ctx context.Context) error {
...
 
+func (sm *stateMachine) Reset(initialState State) {
+	sm.mu.Lock()
+	defer sm.mu.Unlock()
+
+	// Reset current state to initial state
+	sm.currentState = initialState
+
+	// Close the existing event channel to stop processing events
+	close(sm.eventChan)
```
dborovcanin, line 101: Is this safe? Do we have a select on the consumer routine to catch this?

dborovcanin (changes requested):

agent/statemachine/state.go
SammyOina: this happens after the consumer is stopped, we cancel context first so this is safe

drasko (commented):
LGTM

PR #429, opened 2025-05-08
Cocos 407 - Linux IMA

### LINKED ISSUE #407: Feature: Bootloader

In the cloud, we will have to use a custom VM image (probably Ubuntu). Still, we need to change the image to measure the components we can verify ourselves, such as the kernel, initial RAM file system, the kernel command line, and the bootloader. For this purpose, we should research and develop a Cocos bootloader.

Research and develop a bootloader for Cocos image for cloud platforms such as GCP and Azure. Because of the firmware and vTPM used by GCP and Azure, we cannot verify the SEV-SNP measurement, but we can use the vTPM to check what has been booted. The Cocos bootloader should extend vTPM PCR registers to include the hashes of the kernel, initramfs, the kernel command line, and the bootloader itself.

Must-have

### Anything else?

_No response_

jovan-djukic:
# What type of PR is this?
This is a feature because it adds the option of downloading and verifying the Linux IMA measurements.

## What does this do?
This feature enables users to download `ascii_runtime_measurements` file from the VM. In addition, it also verifies the contents of that file using the SHA1 value of TPMs PCR10.

- Resolves #407 

Firmware, bootloader and the kernel are already measured, so the only part of the boot process that isn't measured is the part that includes additional OS modules and such. Linux IMA can be used to solve this.

No. However it was tested successfully on the DELL server.

No.

### Notes

Linux IMA is not enabled by default. To enabled it do the following:

1. Update kernel command line parameters to include `ima_policy=tcb`
2. Update grub
3. Reboot

For this feature to work SHA1 bank of TPM must be enabled. On GCP it is enabled by default. If not, the bank can allocated with the help of tpm2-tools ([tpm2_pcrallocate](https://github.com/tpm2-software/tpm2-tools/blob/master/man/tpm2_pcrallocate.1.md)). On Azure it is also enabled by default. However, If it isn't there is not much that can be done since, unlike GCP which uses vTPM, Azure uses HW TPM which can only be modified through firmware which we don't have access to.

On our DELL server we need to use `swtpm`. SHA1 bank can be enabled through the IGVM file. However, the IGVM must be created with two changes that both impact EDK2:

1. In file SecuirtyPkg/Tcg/Tcg2Dxe/Tcg2Dxe.inf, section [Pcd] needs to be appended with:    `gEfiSecurityPkgTokenSpaceGuid.PcdTpm2HashMask`
4. In file OvmfPkg/OmfPkgX64.dsc, section [PcdsDynamicDefault] needs to be appended with:   `gEfiSecurityPkgTokenSpaceGuid.PcdTpm2HashMask|0x0000000C`

The rest of the compilation process is the same.


drasko:
Please name this PR properly, according to our rules (change title to `COCOS-407 <Some Title>`)

danko-miladinovic (approved):
LGTM

pkg/sdk/agent.go
```diff
@@ -27,15 +27,17 @@ type SDK interface {
 	Data(ctx context.Context, dataset *os.File, filename string, privKey any) error
 	Result(ctx context.Context, privKey any, resultFile *os.File) error
 	Attestation(ctx context.Context, reportData [size64]byte, nonce [size32]byte, attType int, attestationFile *os.File) error
+	IMAMeasurements(ctx context.Context, privKey any, resultFile *os.File) ([]byte, error)
```
SammyOina, line 30: we can leave this unauthenticated as we do in attestation. no need for private key

pkg/sdk/agent.go
```diff
@@ -186,3 +188,38 @@ func generateMetadata(userID string, privateKey crypto.PrivateKey) (metadata.MD,
 	kv[auth.SignatureMetadataKey] = base64.StdEncoding.EncodeToString(signature)
 	return metadata.New(kv), nil
 }
+
+func (sdk *agentSDK) IMAMeasurements(ctx context.Context, privKey any, resultFile *os.File) ([]byte, error) {
+	request := &agent.IMAMeasurementsRequest{}
+
+	md, err := generateMetadata(string(auth.ConsumerRole), privKey)
```
SammyOina, line 195: thid consumer role is part of the computation manifest so it should not be included here

hal/cloud/qemu.sh
```diff
@@ -111,7 +111,8 @@ construct_qemu_args() {
     fi
 
     args+=("-monitor" "$MONITOR")
-    args+=("-no-reboot")
+    # commented so the VM can be rebooted and Linux IMA enabled
+    # args+=("-no-reboot")
```
SammyOina, line 115: if we don't need it erase the commented code

hal/cloud/config.yaml
```diff
@@ -233,8 +233,39 @@ write_files:
       
       echo "algo_user setup completed successfully"
     permissions: "0755"
+    
+  # IMA setup for first boot
+  - path: /cocos_init/linux_ima_init.sh
```
SammyOina, line 238: please open a pr on prism as well adding this change here https://github.com/ultravioletrs/prism/blob/main/docker/providers/cloud-init/config.yml

SammyOina (changes requested):

SammyOina (approved):

agent/api/grpc/server.go
```diff
@@ -235,3 +241,57 @@ func (s *grpcServer) Attestation(req *agent.AttestationRequest, stream agent.Age
...
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return status.Error(codes.Internal, err.Error())
+		}
+
+		if err := stream.Send(&agent.IMAMeasurementsResponse{File: imaMeasurementsResultBuffer[:iman], Pcr10: pcr10ResultBuffer[:pcrn]}); err != nil {
+			return status.Error(codes.Internal, err.Error())
+		}
```
dborovcanin, line 293: Since `rr.File` and `rr.Pcr10` are byte slices, this whole block can be replaced by:
```go
if err := stream.Send(&agent.IMAMeasurementsResponse{File: rr.File, Pcr10: rr.Pcr10}); err != nil {
	return status.Error(codes.Internal, err.Error())
}
```

dborovcanin (changes requested):

agent/api/grpc/server.go
SammyOina: I think this is to read and send in chunks, but this should only be the case if they exceed 4MB in size, the default gRPC limit

dborovcanin: Got it. In that case, this does make sense but requires some adjustments - 1) shorter, more Go-idiomatic names and 2) proper handling of EOF. If either error is EOF - this code will skip both last chunks.

dborovcanin: We also need some retires to `Send` if there are no any implicit ones done by the stream.

SammyOina: we can add abstractions here for sending chunks from agent

dborovcanin: If this is the only place we use it, there is no need to abstract anything.  I guess it's impractical due to the way we handle gRPC server responses (we're always creating new types).

PR #452, opened 2025-06-05
NOISSUE - Fix context handling to fix failing TTL application

SammyOina:
# What type of PR is this?

## What does this do?
This PR fixes the context handling in the manager service, ensuring that the correct context is used for TTL cancellation in CreateVM and throughout the service lifecycle.

### DIFF
--- a/manager/service.go
+++ b/manager/service.go
@@ -99,12 +99,13 @@ type managerService struct {
 	portRangeMax                int
 	persistence                 qemu.Persistence
 	eosVersion                  string
+	ctx                         context.Context
 }
 
 var _ Service = (*managerService)(nil)
 
 // New instantiates the manager service implementation.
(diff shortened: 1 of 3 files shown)


WashingtonKK (approved):

manager/service.go
```diff
@@ -239,7 +241,7 @@ func (ms *managerService) CreateVM(ctx context.Context, req *CreateReq) (string,
 				if err := ms.RemoveVM(ctx, id); err != nil {
 					ms.logger.Error("Failed to remove VM after TTL", "error", err)
 				}
-			case <-ctx.Done():
+			case <-ms.ctx.Done():
```
dborovcanin, line 244: Context is usually not part of the structure, but a param to a function. I.e. context provides an execution context, usually not the whole "world" context. It indicates design problems. Let's merge, but this should be considered in refactoring tasks, so I'll create a ticket out of this comment.

SammyOina: We cannot use context from the function here since it's the grpc request context and will be cancelled when the function returns. But we still need context since we spin up this go routine and don't want to leave it dangling if manager service shuts down before the goroutine returns

Issue #439, opened 2025-05-26
Feature: TDX attestation support


danko-miladinovic:
### Is your feature request related to a problem? Please describe.

TDX attestation support needs to be added to Cocos. Verification and validation, along with aTLS support needs to be added to the CLI and the Agent.

### Describe the feature you are requesting, as well as the possible use case(s) for it.

The changes that need to be added are:
- Verification and validation (CLI)
- aTLS support

### Indicate the importance of this feature to you.

Must-have

SammyOina:
closed by #459

PR #459, opened 2025-06-13
COCOS-439 - TDX Attestation support

danko-miladinovic:
# What type of PR is this?

This is a feature because it adds support for attestion in:
- CLI (verification)
- Agent (fetch attestation report)
- TLS (aTLS)
- Manager (support for attestation policy)

## What does this do?

This PR adds support in Agent, CLI, and Manager for TDX attestation. It enables the Agent to fetch the attestation report from within the CVM (TD) and use it for aTLS.

- Resolves issue #439 

Tests will be added.

Documentation will be added.


pkg/attestation/quoteprovider/sev_test.go
```diff
@@ -70,145 +67,3 @@ func TestFillInAttestationLocal(t *testing.T) {
 		})
 	}
 }
-
-func TestVerifyAttestationReportSuccess(t *testing.T) {
```
SammyOina, line 74: add tests in a follow up pr

danko-miladinovic: Noted.

pkg/attestation/quoteprovider/tdx.go
```diff
@@ -0,0 +1,142 @@
+// Copyright (c) Ultraviolet
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build !embed
+// +build !embed
+
+package quoteprovider
```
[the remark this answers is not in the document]
danko-miladinovic: This is still called the quoteprovider package because the quoteprovider package is the package that provides functions that fetch and verify the TEE attestation reports. But I guess it makes more sense to call it tdx package. I will change this.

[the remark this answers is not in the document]
danko-miladinovic: Changed to tdx package.

Makefile
```diff
@@ -16,7 +16,7 @@ IGVM_BUILD_SCRIPT := ./scripts/igvmmeasure/igvm.sh
 
 define compile_service
 	CGO_ENABLED=$(CGO_ENABLED) GOOS=$(GOOS) GOARCH=$(GOARCH) GOARM=$(GOARM) \
-	go build -ldflags "-s -w \
+	go build -ldflags "-linkmode external -extldflags '-static' -s -w \
```
SammyOina, line 19: Let's do this in a separate PR, and ideally it should include only the agent, or if for all services we need to provide detailed docs on the dependencies or development environment required to build the services

SammyOina (changes requested):

PR #521, opened 2025-09-17
NOISSUE - Refactor http and grpc clients for reusability with Cube

SammyOina:
# What type of PR is this?

## What does this do?
- New Features
  - Unified TLS/mTLS/Attested TLS for clients, with clear security reporting.
  - New HTTP client with env-based configuration and optional ATLS.
  - New HTTP server supporting TLS, mTLS, and Attested TLS with graceful shutdown.

- Refactor
  - Migrated server/gRPC components to public packages; base config renamed to Config.
  - gRPC client now uses shared TLS loaders for simpler, consistent setup.

- Documentation
  - Updated clients package description to reflect HTTP/gRPC client functionality.

- Related Issue #
- Resolves #


dborovcanin (changes requested):
These packages seem to be inadequately positioned/named. First, we have `pkg/http/client.go` with `httpclient` package name. Also, we have `pkg/clients/clients.go` that has no clients in it and handles (a)TLS config and also is being imported by this other package. Please refactor this to make more sense, it will significantly improve readability and quality.

dborovcanin (approved):

Issue #487, opened 2025-07-30
Feature: Refactor attestation package


SammyOina:
### Is your feature request related to a problem? Please describe.

package quoteprovider has redundant functions compared to sev and tdx packages. we need to make the package more aligned and have clear interfaces exposed for verifier and providers 

### Describe the feature you are requesting, as well as the possible use case(s) for it.

we need a refactor

### Indicate the importance of this feature to you.

Must-have

PR #522, opened 2025-09-18
NOISSUE - Refactor aTLS and gRPC server to use CertificateProvider interface

SammyOina:
# What type of PR is this?

## What does this do?
- Removed unused test cases and mock dependencies in atls_test.go.
- Updated TestGetPlatformVerifier to use CertificateVerifier struct.
- Introduced CertificateProvider interface for better abstraction in TLS handling.
- Refactored gRPC server to accept CertificateProvider and configure TLS accordingly.
- Simplified TLS configuration logic in both gRPC and HTTP servers.
- Removed unnecessary parameters from server initialization in tests and main function.
- Enhanced logging for TLS configurations.

- Related Issue #487

### DIFF
--- a/pkg/atls/attestation_provider.go
+++ b/pkg/atls/attestation_provider.go
@@ -7,17 +7,14 @@ import (
 	"fmt"
 
 	"github.com/ultravioletrs/cocos/pkg/attestation"
-	"github.com/ultravioletrs/cocos/pkg/attestation/azure"
-	"github.com/ultravioletrs/cocos/pkg/attestation/tdx"
-	"github.com/ultravioletrs/cocos/pkg/attestation/vtpm"
 	"golang.org/x/crypto/sha3"
 )
 
 // AttestationProvider defines the interface for platform attestation operations.
 type AttestationProvider interface {
-	GetAttestation(pubKey []byte, nonce []byte) ([]byte, error)
-	GetOID() asn1.ObjectIdentifier
-	GetPlatformType() attestation.PlatformType
+	Attest(pubKey []byte, nonce []byte) ([]byte, error)
+	OID() asn1.ObjectIdentifier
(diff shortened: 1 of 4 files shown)


pkg/atls/attestation_provider.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/ultravioletrs/cocos/pkg/attestation"
+	"github.com/ultravioletrs/cocos/pkg/attestation/azure"
+	"github.com/ultravioletrs/cocos/pkg/attestation/tdx"
+	"github.com/ultravioletrs/cocos/pkg/attestation/vtpm"
+	"golang.org/x/crypto/sha3"
+)
+
+// AttestationProvider defines the interface for platform attestation operations.
+type AttestationProvider interface {
+	GetAttestation(pubKey []byte, nonce []byte) ([]byte, error)
```
dborovcanin, line 18: Let's use simply `Attest(pubKey, nonce []byte) ([]byte, error)` signature. Get is usually "getter" or "local", while this fetch is less specific. Do the same for others, no need for `Get` prefix.

pkg/atls/attestation_provider.go
```diff
@@ -0,0 +1,103 @@
...
+	case oid.Equal(SNPvTPMOID):
+		return attestation.SNPvTPM, nil
+	case oid.Equal(AzureOID):
+		return attestation.Azure, nil
+	case oid.Equal(TDXOID):
+		return attestation.TDX, nil
+	default:
+		return 0, fmt.Errorf("unsupported OID: %v", oid)
+	}
+}
```
dborovcanin, line 82: This is used only in cert provider, so no need to have it here.

dborovcanin (changes requested):

PR #558, opened 2025-12-15
NOISSUE - Introduce a dedicated attestation service and refactor agent to use its gRPC client

SammyOina:
# What type of PR is this?

#### Current Architecture Changes
The `cocos-agent` has been refactored to separate the hardware attestation logic into a dedicated component: `attestation-service`.

1.  **Attestation Service**: A standalone service running as `attestation-service` systemd unit.
    *   **Responsibility**: Interacts directly with CVM hardware (TPM/TDX/SNP) to generate attestation quotes and fetch Azure tokens.
    *   **Interface**: Exposes a gRPC interface over a Unix Domain Socket (`/run/cocos/attestation.sock`).
    *   **Security**: Runs with privileges required for hardware access, isolated from the agent's complex workload logic. This follows the principle of least privilege.

2.  **Agent (Coordinator)**: The main `cocos-agent` process.
    *   **Responsibility**: Coordinations computations across the cluster. Managing workloads, verifying manifests, and communicating with the Manager.
    *   **Attestation**: No longer interacts directly with `/dev/tpm` or `/dev/sev`. Instead, it connects to the `attestation-service` via gRPC to request attestation reports.
    *   **Benefit**: Reduced attack surface. If the agent is compromised by a malicious workload, it cannot directly manipulate the attestation hardware.

#### Future Component Splits
To further enhance security and modularity, the following components are proposed to be split from the monolithic agent:

##### 1. Egress Scrutinizer / Networking Proxy
*   **Goal**: Control and monitor all network traffic leaving the CVM.
*   **Implementation**: A separate process or sidecar acting as a transparent proxy.
*   **Function**: Enforce allow-lists for network connections, logging all egress traffic. The agent would route traffic through this component.

##### 2. Computation Runner / Workload Executor
*   **Goal**: Isolate the execution of user algorithms (Python, Binary, Wasm).
*   **Implementation**: A "sandbox" or runner service.
*   **Function**: The agent would submit the algorithm and data specifics to the Runner. The Runner executes the code in a highly restricted environment (e.g., using `seccomp`, `namespaces`, or nested VMs/containers) and returns only the result. This protects the agent (coordinator) from crashes or exploits in user code.

##### 3. Log/Event Forwarder
*   **Goal**: Robust and secure log exfiltration.
*   **Implementation**: A dedicated logging sidecar.
*   **Function**: Buffers logs and events from Agent and other components, ensuring guaranteed delivery to the Manager/Logging service, handling retries and encryption independent of the main agent logic.

#### Diagram (Simplied)

```mermaid
graph TD
    Hardware[Hardware TPM/TDX/SNP]
    AttSock[Attestation Socket /run/cocos/attestation.sock]
    
    subgraph PrivilegedLayer["Privileged Layer"]
        AttSvc[Attestation Service]
    end
    
    subgraph WorkloadLayer["Workload Layer"]
        Agent[Agent Coordinator]
    end
    
    AttSvc -->|Controls| Hardware
    AttSvc -->|Listens| AttSock
    Agent -->|Connects| AttSock
```

- Related Issue #
- Resolves #

### DIFF
--- a/pkg/clients/grpc/attestation/client.go
+++ b/pkg/clients/grpc/attestation/client.go
@@ -63,7 +63,7 @@ func (c *client) GetAttestation(ctx context.Context, reportData [64]byte, nonce 
 		PlatformType: platformType,
 	}
 
-	resp, err := c.client.GetAttestation(ctx, req)
+	resp, err := c.client.FetchAttestation(ctx, req)
 	if err != nil {
 		return nil, err
 	}
@@ -79,7 +79,7 @@ func (c *client) GetAzureToken(ctx context.Context, nonce [32]byte) ([]byte, err
 		Nonce: nonce[:],
 	}
 
-	resp, err := c.client.GetAzureToken(ctx, req)
+	resp, err := c.client.FetchAzureToken(ctx, req)
 	if err != nil {
 		return nil, err
 	}


cmd/attestation-service/main.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	dir := socketPath[:len(socketPath)-len("/attestation.sock")]
+	if err := os.MkdirAll(dir, 0o755); err != nil {
+		logger.Error(fmt.Sprintf("failed to create socket directory: %s", err))
+		exitCode = 1
+		return
+	}
+
+	lis, err := net.Listen("unix", socketPath)
```
dborovcanin: Use `l, err := net.Listen...`, it's more common.

cmd/attestation-service/main.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	if err != nil {
+		return nil, err
+	}
+
+	return &attestationpb.AttestationResponse{Quote: quote}, nil
+}
+
+func (s *service) GetAzureToken(ctx context.Context, req *attestationpb.AzureTokenRequest) (*attestationpb.AzureTokenResponse, error) {
```
dborovcanin: Why do we need this? Is it specific for Azure flow? If this is common for all the cloud providers, can we make the API generic and let service decide which token to return?

SammyOina: It is specific to azure

PR #590, opened 2026-04-22
NOISSUE - Implement extensible resource downloader framework with support for S3, GCS, and OCI sources

SammyOina:
# What type of PR is this?

- Related Issue #
- Resolves #

### DIFF
--- a/agent/service.go
+++ b/agent/service.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"log/slog"
 	"net/http"
+	"net/url"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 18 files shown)


agent/service.go
```diff
@@ -548,28 +563,157 @@ type DecryptedResource struct {
...
+		return resource.SourceTypeS3
+	case strings.HasPrefix(url, "gs://"):
+		return resource.SourceTypeGCS
+	case strings.HasPrefix(url, "https://"):
+		return resource.SourceTypeHTTPS
+	case strings.HasPrefix(url, "http://"):
+		return resource.SourceTypeHTTP
+	default:
+		return ""
+	}
```
smithjilks, line 627: It would be more effective to use scheme instead of prefix matching.
Something like

u, err := urlpkg.Parse(url)
if err != nil {
    return ""
}

smithjilks (changes requested):

agent/service.go
```diff
@@ -548,28 +563,157 @@ type DecryptedResource struct {
...
+		return as.downloadAndDecryptGenericResource(ctx, source, sourceType, resourceType)
 	default:
 		return nil, fmt.Errorf("unsupported source type: %s", sourceType)
 	}
 }
 
+// inferSourceType infers the resource source type from the URL scheme.
+func inferSourceType(url string) string {
+	switch {
+	case strings.HasPrefix(url, "docker://"), strings.HasPrefix(url, "oci:"):
```
smithjilks: We may need to handle a scenario where the user may enter oci without a scheme.

smithjilks (changes requested):

smithjilks (approved):
LGTM

dborovcanin (approved):

Issue #591, opened 2026-04-22
Feature: Add support for NVIDIA CC GPU


jovan-djukic:
### Is your feature request related to a problem? Please describe.

Add support for NVIDIA CC GPU attestation

### Describe the feature you are requesting, as well as the possible use case(s) for it.

Add support for NVIDIA CC GPU attestation

### Indicate the importance of this feature to you.

Must-have

PR #592, opened 2026-04-22
COCOS-591: Add support for GPU CC attestation

jovan-djukic:
# What type of PR is this?
This is a feature because it adds the following functionality: GPU CC attestation. Its related to #591. In order for this feature to work the following conditions need to be met:

| Component | Version | Source |
|---|---|---|
| Host kernel | 6.11.0+ | `coconut-svsm/linux` branch `svsm` |
| Guest kernel | 6.17.0+ | `coconut-svsm/linux` branch `svsm-planes-v6.17` |
| QEMU | 10.1.0-based | `coconut-svsm/qemu` branch `svsm-igvm` |
| COCONUT-SVSM | v2026.02-devel-39-g797c118c | `coconut-svsm/svsm` |
| OVMF | Custom EDK2 | `coconut-svsm/edk2` branch `svsm`, built with `-D TPM2_ENABLE` |
| IGVM library | From `microsoft/igvm` | Installed to `~/igvminst` |
| NVIDIA driver | 570.211.01 | DKMS, built with GCC 14 |
| Guest GCC | 14.x | `ppa:ubuntu-toolchain-r/test` |

The attestion service uses a rust helper to retrieve and verify the attestation report. The path to the helper is passed through an environment variable `ATTESTATION_GPU_HELPER_PATH`.


cmd/attestation-service/gpu.go
```diff
@@ -0,0 +1,76 @@
...
+			// attached, or the helper cannot collect evidence, we continue with
+			// the root CPU/TEE attestation instead of failing the whole request.
+			s.logger.Warn(fmt.Sprintf("[ATTESTATION-SERVICE] Skipping optional GPU evidence collection: %s", err))
+			return opts, nil
+		}
+
+		s.logger.Info(fmt.Sprintf("[ATTESTATION-SERVICE] Collected GPU evidence: format=%s bytes=%d",
+			evidence.EvidenceFormat, len(evidence.RawEvidence)))
+
+		opts = append(opts, eat.WithGPU(&eat.GPUExtensions{
```
SammyOina, line 45: pkg/atls/evidence_verifier.go recomputes the expected GPU nonce from the session nonce, so persisting evidence.Nonce here makes the claim depend on the collector/helper echoing that value back correctly. If that round-trip ever diverges, the service will mint claims that can never verify. Use the gpuNonce you derived locally instead of trusting the helper response for this field.

Makefile
```diff
@@ -23,14 +36,44 @@ define compile_service
 	-o ${BUILD_DIR}/cocos-$(1) ./cmd/$(1)
 endef
 
-.PHONY: all $(SERVICES) install clean
+NVIDIA_ATTESTATION_HELPER_CARGO_ENV = $(if $(filter 1,$(NVAT_USE_SYSTEM_LIB)),NVAT_USE_SYSTEM_LIB=1,)
+NVIDIA_ATTESTATION_HELPER_RUSTFLAGS = $(strip $(RUSTFLAGS) $(if $(filter 1,$(NVAT_USE_SYSTEM_LIB)),,-C link-arg=-Wl,-rpath,$$ORIGIN/lib))
 
-all: $(SERVICES)
+.PHONY: all $(SERVICES) $(NVIDIA_ATTESTATION_HELPER) nvidia-attestation-helper-prereqs install clean
```
SammyOina, line 42: make all now hard-requires NVAT setup, which is a build regression.

Default builds will fail on environments without NVAT headers/checkout, even when GPU attestation is not needed.

pkg/atls/evidence_verifier.go
```diff
@@ -90,3 +112,39 @@ func platformVerifier(platformType cocosattestation.PlatformType) (cocosattestat
...
+
+	return gpu.NewVerifier(binaryPath, timeout)
+}
+
+func (v *policyEvidenceVerifier) verifyGPUEvidence(claims *eat.EATClaims, manifest *corim.UnsignedCorim) error {
+	if len(claims.GPUExtensions.EvidenceJSON) == 0 {
+		return fmt.Errorf("atls: gpu evidence is empty")
+	}
+
+	expectedNonce := sha256.Sum256(append(append([]byte(nil), claims.Nonce...), []byte(":gpu")...))
```
SammyOina, line 142: Right now the session binding check only covers claims.GPUExtensions.Nonce. The raw GPU evidence is then verified using the nonce embedded in claims.GPUExtensions.EvidenceJSON, and those two values are never compared. That leaves a replay gap: a stale but self-consistent GPU evidence blob can be paired with a rewritten outer nonce and still pass helper verification. Reject the claim unless the nonce inside EvidenceJSON matches expectedNonce.

pkg/attestation/eat/eat.go
```diff
@@ -129,6 +160,12 @@ func NewEATClaims(report []byte, nonce []byte, platformType attestation.Platform
 		return nil, err
 	}
 
+	for _, opt := range opts {
```
SammyOina, line 163: Guard nil ClaimsOption values to prevent panic. A call like NewEATClaims(..., nil) will panic at option invocation.

pkg/attestation/gpu/verifier.go
```diff
@@ -0,0 +1,127 @@
...
+
+	var resp helperResponse
+	if err := json.Unmarshal(stdout.Bytes(), &resp); err != nil {
+		return fmt.Errorf("failed to decode GPU verifier response: %w", err)
+	}
+	if len(resp.ClaimsJSON) == 0 {
+		return fmt.Errorf("gpu verifier response did not contain claims_json")
+	}
+
+	// NVIDIA attestation currently performs its own evidence-policy appraisal
```
SammyOina: pkg/atls/evidence_verifier.go now passes the deployment policy into VerifyWithCoRIM, but this implementation discards it and accepts whatever the helper’s built-in NVIDIA checks allow. That makes GPU appraisal independent of the configured CoRIM and can admit hardware that the operator policy did not authorize. Please either appraise the returned GPU claims against manifest here or fail closed until manifest enforcement exists.

tools/nvidia-attestation-helper/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "nvidia-attestation-helper"
+version = "0.1.0"
+edition = "2021"
+license = "Apache-2.0"
+
+[dependencies]
+anyhow = "1"
+nv-attestation-sdk = { git = "https://github.com/NVIDIA/attestation-sdk", branch = "main" }
```
SammyOina: Using branch = "main" makes builds non-deterministic and can break unexpectedly as upstream changes. Without a Cargo.lock file, this dependency will pull the latest main branch commit on every build.

SammyOina (changes requested):

pkg/attestation/gpu/verifier.go
```diff
@@ -0,0 +1,228 @@
...
+	if len(resp.ClaimsJSON) == 0 {
+		return fmt.Errorf("gpu verifier response did not contain claims_json")
+	}
+
+	var deviceClaims map[string]gpuDeviceClaims
+	if err := json.Unmarshal(resp.ClaimsJSON, &deviceClaims); err != nil {
+		return fmt.Errorf("gpu: failed to parse claims JSON: %w", err)
+	}
+
+	return appraiseGPUClaims(deviceClaims, manifest)
```
SammyOina, line 132: {} passes the raw len(resp.ClaimsJSON) check, unmarshals into an empty map, and then appraiseGPUClaims returns success without verifying any device. Treat len(deviceClaims) == 0 as an error before appraisal.

Makefile
```diff
@@ -23,14 +36,44 @@ define compile_service
...
+	mkdir -p $(BUILD_DIR)
+
+$(SERVICES): | $(BUILD_DIR)
 	$(call compile_service,$@)
 	@if [ "$@" = "cli" ] || [ "$@" = "manager" ]; then $(MAKE) build-igvm; fi
 
+nvidia-attestation-helper-prereqs:
+ifeq ($(filter 1,$(NVAT_USE_SYSTEM_LIB)),1)
+	@test -f $(NVAT_SYSTEM_HEADER) || \
+		( echo "Missing $(NVAT_SYSTEM_HEADER). Install the NVAT development package or run without NVAT_USE_SYSTEM_LIB=1."; exit 1 )
```
SammyOina, line 56: This branch only validates nvat.h. If the header is installed but libnvat.so.1 is not available on the link/runtime path, prereqs pass and the failure gets deferred to Cargo with a much noisier linker error. Validate the shared library in this branch as well so NVAT_USE_SYSTEM_LIB=1 fails early and clearly.

SammyOina (changes requested):

SammyOina (approved):

PR #597, opened 2026-06-04
NOISSUE - Add initial implementation of COCOS SPI components

SammyOina:
# What type of PR is this?

## What does this do?
- Introduced ComputationJobStore interface for managing computation jobs.
- Added ComputationOrchestrator interface for orchestrating computation requests.
- Created RemoteAssetFetcher interface for fetching assets from remote providers.
- Defined model classes: AgentConfig, AlgorithmSpec, AssetSource, ComputationJob, ComputationRequest, ComputationUnit, ComputeManifest, and DatasetSpec for computation job management.
- Added Gradle configuration files including gradle.properties and libs.versions.toml for dependency management.
- Included Gradle wrapper files for consistent build environment.
- Updated settings.gradle.kts to include COCOS modules.

- Related Issue #
- Resolves #


drasko:
I do not like the way this is solved - can we consider creating a thin shim in Go, a proxy microservice that will do the actual attestation? I would like to avoid changing EDC things and maintaining their complex port. Either get our stuff merged with them (which I doubt that will acceptedm, as Cocos might be specific), or leverage on generic components (like KMS from CoCo), but please do not change and fork these huge projects without the strategy how this can be maintained.

